PR #36, opened 2022-07-25
chore: updating npm packages

kupermind:
- Updating npm packages.


contracts/test/GnosisSafeABICreator.sol
```diff
@@ -3,5 +3,4 @@ pragma solidity ^0.8.15;
 
 // Getting ABIs for the Gnosis Safe master copy and proxy contracts
 import "@gnosis.pm/safe-contracts/contracts/GnosisSafe.sol";
-import "@gnosis.pm/safe-contracts/contracts/proxies/GnosisSafeProxy.sol";
```
kupermind, line 6: This contract is not used even on the ABI level after we switched to interface usage of Gnosis Safe

hardhat.config.js
```diff
@@ -24,7 +23,7 @@ const ETHERSCAN_API_KEY = process.env.ETHERSCAN_API_KEY;
 
 module.exports = {
     networks: {
-        ganache: {
+        local: {
```
kupermind, line 26: This would be more appropriate for our usage

DavidMinarsch (approved):

mariapiamo (approved):

PR #37, opened 2022-08-17
fix: multisig implementation

kupermind:
- Fixing a bug in the Gnosis Safe multisig implementation;
- Adding corresponding tests.


contracts/multisigs/GnosisSafeMultisig.sol
```diff
@@ -14,11 +14,18 @@ interface IGnosisSafeProxyFactory {
...
+error IncorrectDataLength(uint256 expected, uint256 provided);
+
 /// @title Gnosis Safe - Smart contract for Gnosis Safe multisig implementation of a generic multisig interface
 /// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
 contract GnosisSafeMultisig {
     // Selector of the Gnosis Safe setup function
-    bytes4 internal constant _GNOSIS_SAFE_SETUP_SELECTOR = 0xb63e800d;
+    bytes4 public constant GNOSIS_SAFE_SETUP_SELECTOR = 0xb63e800d;
+    // Default data size to be parsed and passed to the Gnosis Safe Factory without payload
+    uint256 public constant DEFAULT_DATA_LENGTH = 144;
```
kupermind, line 28: Adding the default data length when the passed data length is different from the zero

contracts/multisigs/GnosisSafeMultisig.sol
```diff
@@ -33,15 +40,22 @@ contract GnosisSafeMultisig {
...
             uint256 payment, uint256 nonce, bytes memory payload)
     {
-        if (data.length > 0) {
-            uint256 dataSize = data.length;
+        uint256 dataLength = data.length;
+        if (dataLength > 0) {
+            // Check for the correct data length
+            if (dataLength < DEFAULT_DATA_LENGTH) {
+                revert IncorrectDataLength(DEFAULT_DATA_LENGTH, dataLength);
+            }
```
kupermind, line 54: If the data is not zero, we need to check for it to be at least 144 bytes length, otherwise we are guaranteed to not parse it correctly

contracts/multisigs/GnosisSafeMultisig.sol
```diff
@@ -51,14 +65,20 @@ contract GnosisSafeMultisig {
...
 
-                // Read the payload data
-                payload := mload(add(data, dataSize))
+            // Read the payload, if provided
+            if (dataLength > DEFAULT_DATA_LENGTH) {
+                uint256 payloadLength = dataLength - DEFAULT_DATA_LENGTH;
+                payload = new bytes(payloadLength);
+                for (uint256 i = 0; i < payloadLength; ++i) {
+                    payload[i] = data[i + DEFAULT_DATA_LENGTH];
+                }
```
kupermind, line 81: Extracting `bytes` payload in assembly is not that straightforward, it is more convenient and user-friendly to do it in plain solidity

contracts/multisigs/GnosisSafeMultisig.sol
```diff
@@ -14,11 +14,18 @@ interface IGnosisSafeProxyFactory {
...
+/// @param expected Expected minimum data length.
+/// @param provided Provided data length.
+error IncorrectDataLength(uint256 expected, uint256 provided);
+
 /// @title Gnosis Safe - Smart contract for Gnosis Safe multisig implementation of a generic multisig interface
 /// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
 contract GnosisSafeMultisig {
     // Selector of the Gnosis Safe setup function
-    bytes4 internal constant _GNOSIS_SAFE_SETUP_SELECTOR = 0xb63e800d;
+    bytes4 public constant GNOSIS_SAFE_SETUP_SELECTOR = 0xb63e800d;
```
kupermind, line 26: No need to leave this `internal`

test/GnosisSafeMultisig.js
```diff
@@ -0,0 +1,139 @@
+/*global describe, context, beforeEach, it*/
+
+const { expect } = require("chai");
+const { ethers } = require("hardhat");
+
+describe("GnosisSafeMultisig", function () {
```
kupermind, line 6: Adding a set of tests for the multisig implementation

contracts/multisigs/GnosisSafeMultisig.sol
DavidMinarsch: What was the error on the lhs?

77ph (approved):

contracts/multisigs/GnosisSafeMultisig.sol
kupermind: The allocation (line 78) is needed before copying the bytes themselves

DavidMinarsch (approved):

mariapiamo (approved):

PR #40, opened 2022-08-22
feat and test: ability to re-deploy the service with the same multisig

kupermind:
- Added the multisig implementation contract that checks the modified multisig ownership and threshold due to a changed number of agent instances, and lets the service deploy with the same multisig from the previous deployment;
- Added tests.


contracts/multisigs/GnosisSafeSameAddressMultisig.sol
```diff
@@ -0,0 +1,91 @@
...
+
+/// @dev Provided incorrect multisig owner.
+/// @param provided Provided owner address.
+error WrongOwner(address provided);
+
+/// @title Gnosis Safe Same Address - Smart contract for Gnosis Safe verification of an already existent multisig address.
+/// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
+contract GnosisSafeSameAddressMultisig {
+    // Default data size to be parsed as an address of a Gnosis Safe multisig
+    uint256 public constant DEFAULT_DATA_LENGTH = 20;
```
kupermind, line 38: All we need is the address of a multisig that was already deployed

contracts/multisigs/GnosisSafeSameAddressMultisig.sol
```diff
@@ -0,0 +1,91 @@
...
+// Gnosis Safe Master Copy interface extracted from the mainnet: https://etherscan.io/address/0xd9db270c1b5e3bd161e8c8503c55ceabee709552#code#F6#L126
+interface IGnosisSafe {
+    /// @dev Gets set of owners.
+    /// @return Set of Safe owners.
+    function getOwners() external view returns (address[] memory);
+
+    /// @dev Gets threshold.
+    /// @return Threshold
+    function getThreshold() external view returns (uint256);
+}
```
kupermind, line 13: Checks through the master copy gnosis safe to get the ownership information about the multisig

contracts/multisigs/GnosisSafeSameAddressMultisig.sol
```diff
@@ -0,0 +1,91 @@
...
+        }
+        uint256 numOwners = owners.length;
+        if (numOwners != checkOwners.length) {
+            revert WrongNumOwners(checkOwners.length, numOwners);
+        }
+        // The owners in the multisig itself are stored in reverse order compared to how they were added.
+        // Thus, the check must be carried out accordingly. Use the following code for reference:
+        // https://etherscan.io/address/0xd9db270c1b5e3bd161e8c8503c55ceabee709552#code#F6#L56
+        for (uint256 i = 0; i < numOwners; ++i) {
+            if (owners[i] != checkOwners[numOwners - i - 1]) {
```
kupermind, line 86: This is an important moment - the sequence of how the new addresses were added

test/GnosisSafeSameAddressMultisig.js
```diff
@@ -0,0 +1,105 @@
...
+                gnosisSafeSameAddressMultisig.create(newOwnerAddresses, initialThreshold, "0x55")
+            ).to.be.revertedWith("IncorrectDataLength");
+
+            const data = AddressZero + "55";
+            await expect(
+                gnosisSafeSameAddressMultisig.create(newOwnerAddresses, initialThreshold, data)
+            ).to.be.revertedWith("IncorrectDataLength");
+        });
+
+        it("Create a multisig and change its owners and threshold", async function () {
```
kupermind, line 48: Verifying the create() function that in reality does the verification of a new multisig setup

test/ServiceRegistry.js
```diff
@@ -818,6 +823,88 @@ describe("ServiceRegistry", function () {
...
+            /// Register agent instance
+            await serviceRegistry.connect(serviceManager).registerAgents(operator, serviceId,
+                [agentInstances[0].address], [agentId], {value: regBond});
+
+            // Whitelist both gnosis multisig implementations
+            await serviceRegistry.changeMultisigPermission(gnosisSafeMultisig.address, true);
+            await serviceRegistry.changeMultisigPermission(gnosisSafeSameAddressMultisig.address, true);
+
+            // Deploy the service and create a multisig and get its address
+            const safe = await serviceRegistry.connect(serviceManager).deploy(owner, serviceId,
```
kupermind, line 857: Deploy the service with one agent instance (the multisig is created)

test/ServiceRegistry.js
```diff
@@ -818,6 +823,88 @@ describe("ServiceRegistry", function () {
...
+            const multisig = await ethers.getContractAt("GnosisSafe", proxyAddress);
+
+            // Terminate a service after some time since there's a need to add agent instances
+            await serviceRegistry.connect(serviceManager).terminate(owner, serviceId);
+
+            // Unbond the agent instance in order to update the service
+            await serviceRegistry.connect(serviceManager).unbond(operator, serviceId);
+
+            // Updating a service
+            await serviceRegistry.connect(serviceManager).update(owner, configHash, [1], [[4, regBond]],
```
kupermind, line 871: Update multisig to have now 4 agent instances

test/ServiceRegistry.js
```diff
@@ -818,6 +823,88 @@ describe("ServiceRegistry", function () {
...
+                newMaxThreshold, serviceId);
+
+            // Activate agent instance registration
+            await serviceRegistry.connect(serviceManager).activateRegistration(owner, serviceId, {value: regDeposit});
+
+            /// Register agent instance
+            await serviceRegistry.connect(serviceManager).registerAgents(operator, serviceId,
```
kupermind, line 878: Register all 4 agent instances with the original agent instance being the leading one

test/ServiceRegistry.js
```diff
@@ -818,6 +823,88 @@ describe("ServiceRegistry", function () {
...
+                [agentInstances[0].address, agentInstances[1].address, agentInstances[2].address, agentInstances[3].address],
+                [agentId, agentId, agentId, agentId], {value: 4 * regBond});
+
+            // Change the existent multisig owners and threshold using the access of a previously used agent instance
+            const safeContracts = require("@gnosis.pm/safe-contracts");
+            // Add agent instances as owners except for the first original one
+            for (let i = 1; i < agentInstances.length; i++) {
```
kupermind, line 885: Modify the original multisig setup using the original agent instance to add more agent instances and change a threshold

test/ServiceRegistry.js
```diff
@@ -818,6 +823,88 @@ describe("ServiceRegistry", function () {
...
+            const nonce = await multisig.nonce();
+            const txHashData = await safeContracts.buildContractCall(multisig, "changeThreshold",
+                [newMaxThreshold], nonce, 0, 0);
+            const signMessageData = await safeContracts.safeSignMessage(agentInstances[0], multisig, txHashData, 0);
+            await safeContracts.executeTx(multisig, txHashData, [signMessageData], 0);
+
+            // Pack the original multisig address
+            const data = ethers.utils.solidityPack(["address"], [multisig.address]);
+            // Redeploy the service using the newly updated multisig (same multisig address)
+            await serviceRegistry.connect(serviceManager).deploy(owner, serviceId, gnosisSafeSameAddressMultisig.address, data);
```
kupermind, line 902: Provide the original multisig address and redeploy the service with a new multisig setup

test/ServiceRegistry.js
DavidMinarsch: It's an interesting requirement. Means we should also update the Manager contract to ensure that the order can be guaranteed. Let's discuss tomorrow.

DavidMinarsch (approved):
LGTM

contracts/multisigs/GnosisSafeSameAddressMultisig.sol
kupermind: Updated that part. Actually, the master copy address is not needed at all, we don't use it except for the interface

contracts/multisigs/GnosisSafeSameAddressMultisig.sol
```diff
@@ -0,0 +1,83 @@
...
+    ) external view returns (address multisig)
+    {
+        // Check for the correct data length
+        if (data.length != DEFAULT_DATA_LENGTH) {
+            revert IncorrectDataLength(DEFAULT_DATA_LENGTH, data.length);
+        }
+
+        // Read the proxy multisig address (20 bytes)
+        assembly {
+            multisig := mload(add(data, DEFAULT_DATA_LENGTH))
```
mariapiamo, line 59: Last time it was mentioned that extracting bytes payload in assembly is not that straightforward, why not do this in plain solidity?

contracts/multisigs/GnosisSafeSameAddressMultisig.sol
```diff
@@ -0,0 +1,83 @@
...
+
+        // Verify the multisig for owners and threshold
+        if (threshold != checkThreshold) {
+            revert WrongThreshold(checkThreshold, threshold);
+        }
+        uint256 numOwners = owners.length;
+        if (numOwners != checkOwners.length) {
+            revert WrongNumOwners(checkOwners.length, numOwners);
+        }
+        // The owners in the multisig itself are stored in reverse order compared to how they were added.
```
mariapiamo, line 74:
```suggestion
        // The owners in the multisig itself are stored in reverse order compared to how they were added:
        // https://etherscan.io/address/0xd9db270c1b5e3bd161e8c8503c55ceabee709552#code#F6#L106
```

contracts/multisigs/GnosisSafeSameAddressMultisig.sol
kupermind: There is no straightforward way to convert `bytes` to `uint256`. It was mentioned that `bytes` into sub-`bytes` is not so trivial, but `bytes` into static-sized variables is straightforward.

mariapiamo (approved):

contracts/multisigs/GnosisSafeSameAddressMultisig.sol
```diff
@@ -0,0 +1,83 @@
...
+
+/// @title Gnosis Safe Same Address - Smart contract for Gnosis Safe verification of an already existent multisig address.
+/// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
+contract GnosisSafeSameAddressMultisig {
+    // Default data size to be parsed as an address of a Gnosis Safe multisig proxy address
+    uint256 public constant DEFAULT_DATA_LENGTH = 20;
+
+    /// @dev Verifies the existent gnosis safe multisig for changed owners and threshold.
+    /// @notice The multisig was supposedly updated before reaching this step such that new multisig is not created.
+    /// @param owners Set of multisig owners.
```
DavidMinarsch, line 42: Please add in comment the expected order of owners!

DavidMinarsch (changes requested):
Tiny requested change - o/w lgtm

contracts/multisigs/GnosisSafeSameAddressMultisig.sol
DavidMinarsch: @kupermind don’t refer to “comments below”. When deployed services like etherscan use the Nat spec for the UI but they won’t show code comments. So the natspec should be self contained.

DavidMinarsch (approved):

PR #41, opened 2022-08-23
feat: L2 service registry

kupermind:
- Adding a contract for a service registry to be deployed on L2;
- Added the testing of L2 service registry contract along with L1 testing, where applicable (for example, subcomponents can't be tested in L2).


contracts/ServiceRegistryL2.sol
```diff
@@ -0,0 +1,878 @@
...
+// This struct is 192 bits in total
+struct AgentInstance {
+    // Address of an agent instance
+    address instance;
+    // Canonical agent Id. This number is limited by the max number of agent Ids (see UnitRegistry contract)
+    uint32 agentId;
+}
+
+/// @title Service Registry L2 - Smart contract for registering services on L2 chains.
+/// @dev Underlying canonical agents and components are not checked for their validity since they are set up on the L1 mainnet.
```
kupermind, line 25: Please check this statement

DavidMinarsch, line 25:
```suggestion
/// @dev Underlying canonical agents and components are not checked for their validity since they are set up on the L1 mainnet. The architecture is optimistic, in the sense that service owners are assumed to reference existing and relevant agents.
```

test/ServiceRegistryL2.js
```diff
@@ -0,0 +1,1167 @@
+/*global describe, context, beforeEach, it*/
+
+const { expect } = require("chai");
+const { ethers } = require("hardhat");
+
+describe("ServiceRegistryL2", function () {
```
DavidMinarsch, line 6: It is a copy and paste I assume. It would be great if it was very easy to see the difference to L1

test/ServiceRegistryL2.js
```diff
@@ -0,0 +1,1167 @@
...
+            ).to.be.revertedWith("WrongArrayLength");
+            await expect(
+                serviceRegistry.connect(serviceManager).create(owner, configHash, [1], [], threshold)
+            ).to.be.revertedWith("WrongArrayLength");
+            await expect(
+                serviceRegistry.connect(serviceManager).create(owner, configHash, [1, 3], [[2, regBond]],
+                    threshold)
+            ).to.be.revertedWith("WrongArrayLength");
+        });
+
```
mariapiamo, line 168: what do we expect to happen when a service is created with a non-existent canonical agent?

contracts/ServiceRegistryL2.sol
mariapiamo, line 25:
```suggestion
/// @dev Underlying canonical agents and components are not checked for their validity since they are set up on the L1 mainnet and because we follow an optimistic design approach ( whether service owners point to the rights underlying canonical agents and components is up to them ) 
```

test/ServiceRegistryL2.js
kupermind: It is not really a copy and paste, there are some function calls related to agent services missing, I didn't find a better way ATM

DavidMinarsch: Ok, so at a min, let's add a very basic check to the CI: `diff L1file L2file` and then assert the output is what we expect (a hardcoded string). That's a bit hacky but adds a lot of transparency on what exactly the diff is between the two versions :)

test/ServiceRegistryL2.js
mariapiamo: Ok, so not sure if it makes sense to add somewhere that whether the service is stuck to the agent instance registration, it is possible that one of the canonical agents is non-existent and if the latter is verified the service can be updated with the right ids.

kupermind: We have stated that at the top of the contract

test/ServiceRegistry.js
```diff
@@ -31,6 +33,8 @@ describe("ServiceRegistry", function () {
     const agentHash2 = "0x" + "9".repeat(64);
     const AddressZero = "0x" + "0".repeat(40);
     const payload = "0x";
+    const serviceRegistryImplementations = ["l1", "l2"];
```
kupermind, line 36: Almost each context do a separate run for L1 and L2 implementation

test/ServiceRegistry.js
```diff
@@ -50,1495 +54,1854 @@ describe("ServiceRegistry", function () {
...
-            ).to.be.revertedWith("ZeroValue");
-
-            await serviceRegistry.setBaseURI("https://localhost2/service/");
-            expect(await serviceRegistry.baseURI()).to.equal("https://localhost2/service/");
+        serviceRegistryImplementations.forEach(async function (serviceRegistryImplementation) {
+            it("Changing owner", async function () {
+                // Choose the service registry implementation: L1 or L2
+                if (serviceRegistryImplementation == "l2") {
+                    serviceRegistry = serviceRegistryL2;
+                }
```
kupermind, line 90: The service registry contract instance is adjusted depending on the L1 or L2 run

test/ServiceRegistry.js
```diff
@@ -50,1495 +54,1854 @@ describe("ServiceRegistry", function () {
...
+
+                const mechManager = signers[3];
+                const serviceManager = signers[4];
+                const owner = signers[5].address;
+                // Components and agents can only be created on L1
+                if (serviceRegistryImplementation == "l1") {
+                    await agentRegistry.changeManager(mechManager.address);
+                    await agentRegistry.connect(mechManager).create(owner, agentHash, [1]);
+                }
```
kupermind, line 276: Essentially components and agents can be created on L1 only

contracts/ServiceRegistry.sol
```diff
@@ -382,7 +382,7 @@ contract ServiceRegistry is GenericRegistry {
 
     /// @dev Registers agent instances.
     /// @param operator Address of the operator.
-    /// @param serviceId Service Id to be updated.
+    /// @param serviceId Service Id to register agent instances for.
```
DavidMinarsch, line 385: revert

DavidMinarsch (changes requested):
Please revert the code comment on the deployed code

DavidMinarsch (approved):

PR #50, opened 2022-09-08
chore: add docker setup

DavidMinarsch:


yarn.lock
```diff
@@ -720,10 +720,10 @@
     "@types/sinon-chai" "^3.2.3"
     "@types/web3" "1.0.19"
 
-"@openzeppelin/contracts@^4.7.1":
-  version "4.7.1"
-  resolved "https://registry.yarnpkg.com/@openzeppelin/contracts/-/contracts-4.7.1.tgz#33d0a857b76b18d313e6bb6ed28cdaf367e90cfe"
-  integrity sha512-UXmAjKARsXORHlHZu5GCD7ZbRKm6nU8UHnbuT/QJJa2JEOEcbvV/X8w/sUk62Sl9VZuuljM1akrZLyAtzUgsxw==
+"@openzeppelin/contracts@^4.7.2":
```
DavidMinarsch, line 723: Can we roll this back to this version please: https://github.com/valory-xyz/autonolas-registries/blob/v1.0.0/yarn.lock#L774

This is the one we use on mainnet. Bumping this here and in previous PRs  can cause problems.

PR #54, opened 2022-11-17
test: redeploy a service with the same multisig owned by a multisig

kupermind:
- Simulating redeployment of a service with the same multisig (parent multisig) owned by the service owner multisig (child multisig);
- Data location inheritance bug in the Safe interface: https://blog.soliditylang.org/2022/05/17/data-location-inheritance-bug/
- Use `DefaultCallbackHandler` instead of the `CompatibilityFallbackHandler` for the service owner multisig to avoid manual patching of gnosis safe files, and yet to have the ERC721 support.

### DIFF
--- a/contracts/test/GnosisSafeABICreator.sol
+++ b/contracts/test/GnosisSafeABICreator.sol
@@ -5,3 +5,5 @@ pragma solidity ^0.8.15;
 import "@gnosis.pm/safe-contracts/contracts/GnosisSafe.sol";
 import "@gnosis.pm/safe-contracts/contracts/proxies/GnosisSafeProxyFactory.sol";
 import "@gnosis.pm/safe-contracts/contracts/libraries/MultiSendCallOnly.sol";
+import "@gnosis.pm/safe-contracts/contracts/handler/CompatibilityFallbackHandler.sol";
+import "@gnosis.pm/safe-contracts/contracts/examples/libraries/SignMessage.sol";
--- a/contracts/test/GnosisSafeABICreator.sol
+++ b/contracts/test/GnosisSafeABICreator.sol
@@ -3,7 +3,6 @@ pragma solidity ^0.8.15;
 
 // Getting ABIs for the Gnosis Safe master copy and proxy contracts
 import "@gnosis.pm/safe-contracts/contracts/GnosisSafe.sol";
+import "@gnosis.pm/safe-contracts/contracts/handler/DefaultCallbackHandler.sol";
 import "@gnosis.pm/safe-contracts/contracts/proxies/GnosisSafeProxyFactory.sol";
 import "@gnosis.pm/safe-contracts/contracts/libraries/MultiSendCallOnly.sol";
-import "@gnosis.pm/safe-contracts/contracts/handler/CompatibilityFallbackHandler.sol";
-import "@gnosis.pm/safe-contracts/contracts/examples/libraries/SignMessage.sol";
--- a/scripts/deployment/deploy_06_gnosis_safe_multisig.js
+++ b/scripts/deployment/deploy_06_gnosis_safe_multisig.js
@@ -11,8 +11,6 @@ async function main() {
     const useLedger = parsedData.useLedger;
     const derivationPath = parsedData.derivationPath;
     const providerName = parsedData.providerName;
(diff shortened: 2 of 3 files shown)


test/ServiceRegistry.js
```diff
@@ -1416,6 +1427,166 @@ describe("ServiceRegistry", function () {
...
+
+                // Create a message data from the multisend transaction
+                const messageData = await multisig.encodeTransactionData(safeTx.to, safeTx.value, safeTx.data,
+                    safeTx.operation, safeTx.safeTxGas, safeTx.baseGas, safeTx.gasPrice, safeTx.gasToken,
+                    safeTx.refundReceiver, nonce);
+
+                // Sign multisend message for the service owner multisig
+                await safeContracts.executeContractCallWithSigners(serviceOwnerMultisig, signMessageLib, "signMessage",
+                    [messageData], [serviceOwnerOwners[0], serviceOwnerOwners[1]], true);
+                // on the front-end something like: await signMessageLib.connect(serviceOwnerMultisig).signMessage(messageData);
```
kupermind, line 1546: Need to figure out how to call this correctly via the frontend.

77ph:
Fix test please.
Other ok.

scripts/deployment/deploy_06_gnosis_safe_multisig.js
```diff
@@ -11,8 +11,6 @@ async function main() {
     const useLedger = parsedData.useLedger;
     const derivationPath = parsedData.derivationPath;
     const providerName = parsedData.providerName;
-    const gnosisSafeAddress = parsedData.gnosisSafeAddress;
-    const gnosisSafeProxyFactoryAddress = parsedData.gnosisSafeProxyFactoryAddress;
```
kupermind, line 15: These are obsolete, the linter complains

contracts/test/GnosisSafeABICreator.sol
```diff
@@ -3,5 +3,6 @@ pragma solidity ^0.8.15;
 
 // Getting ABIs for the Gnosis Safe master copy and proxy contracts
 import "@gnosis.pm/safe-contracts/contracts/GnosisSafe.sol";
+import "@gnosis.pm/safe-contracts/contracts/handler/DefaultCallbackHandler.sol";
```
kupermind, line 6: DefaultCallbackHandler is needed since the ERC721 support is required for the service owner multisig.

mariapiamo (approved):

77ph (approved):

contracts/test/GnosisSafeABICreator.sol
DavidMinarsch: @kupermind make sure that the fronend makes this known to the user. Maybe we can add a check and error message if it’s not set.

DavidMinarsch (approved):

contracts/test/GnosisSafeABICreator.sol
kupermind: All gnosis safe contracts created via Gnosis Safe App have at least the `DefaultCallbackHandler` by default. Those of the latest version 1.3 have the `CompatibilityFallbackHandler` by default. The only way to check that is to read the storage of the proxy contract.

We can have the message when we identify that it's a safe that wants to create a component / agent / service, also we can have an additional check for the proxy storage fallback handler slot.

kupermind: Then get the address by a slot:
`FALLBACK_HANDLER_STORAGE_SLOT = 0x6c9a6c4a39284e37ed1cf53d337577d14212a4870fb976a4366c693b939918d5`

Then read the value of the pure `onERC721Received()` function

kupermind: python:
`slot = w3_mainnet.eth.getStorageAt(gnosis_safe_address_reference,FALLBACK_HANDLER_STORAGE_SLOT).hex()`

DavidMinarsch: Suggest you open an issue for Mohan

kupermind: It's done already in 2 tickets )

PR #55, opened 2022-11-28
chore: checking for the passed service config hash in docker

kupermind:
- Checking for the passed service config hash in docker.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
-echo "Using SERVICE_CONFIG_HASH = $SERVICE_CONFIG_HASH"
-echo $(jq --arg SERVICE_CONFIG_HASH "$SERVICE_CONFIG_HASH" '.serviceRegistry.configHashes = [ $SERVICE_CONFIG_HASH ]' snapshot.json) > snapshot.json
-yarn hardhat node
+if [[ -n "$SERVICE_CONFIG_HASH" ]]; then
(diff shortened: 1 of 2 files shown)


entrypoint.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
-echo "Using SERVICE_CONFIG_HASH = $SERVICE_CONFIG_HASH"
-echo $(jq --arg SERVICE_CONFIG_HASH "$SERVICE_CONFIG_HASH" '.serviceRegistry.configHashes = [ $SERVICE_CONFIG_HASH ]' snapshot.json) > snapshot.json
-yarn hardhat node
+if [[ -n "$SERVICE_CONFIG_HASH" ]]; then
+  echo "Using SERVICE_CONFIG_HASH = $SERVICE_CONFIG_HASH"
+  echo $(jq --arg SERVICE_CONFIG_HASH "$SERVICE_CONFIG_HASH" '.serviceRegistry.configHashes = [ $SERVICE_CONFIG_HASH ]' snapshot.json) > snapshot.json
+  yarn hardhat node
```
angrybayblade, line 6: Move this out of the `if` block

kupermind: You mean the last line? If we run it with the empty config hash, it will just fail, since the first service registry creation function will check for the hash length and revert.

angrybayblade: It'll still work since we will still hash in the config hash in `snapshot.json`, I tried this locally and it worked

angrybayblade (approved):

PR #60, opened 2023-01-23
Add release flow for docker images

angrybayblade:


8ball030:
whilst this lookls good to me, i am not sure until it runs, you have built the images?

Post successful run i would pull the images from the registry and then ensure they contain what you expect.

8ball030 (approved):

PR #62, opened 2023-01-25
This PR fixes missing `snapshot.json` file issue in the docker image

angrybayblade:


deploy/0_deploy.js
```diff
@@ -5,8 +5,7 @@ const { expect } = require("chai");
 module.exports = async () => {
     // Read configs from the JSON file
     const fs = require("fs");
-    // Copy this file from scripts/mainnet_snapshot.json or construct one following the JSON defined structure
-    const snapshotFile = "snapshot.json";
+    const snapshotFile = "/base/scripts/mainnet_snapshot.json";
```
angrybayblade, line 8: We can use the snapshot file directly from the dir we copied in the `/base` dir

kupermind (approved):

DavidMinarsch (approved):

PR #68, opened 2023-04-05
feat: safe guard proof of concept

kupermind:
- Introducing the Safe guard concept;
- The guard is set during the multisig setup and correctly written into its guard slot;
- The guard checks the multisig transaction execution for a specific condition.

### DIFF
--- a/contracts/multisigs/SafeGuard.sol
+++ b/contracts/multisigs/SafeGuard.sol
@@ -1,0 +1,90 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.19;
+
+import "hardhat/console.sol";
+
+/// @dev Only `owner` has a privilege, but the `sender` was provided.
+/// @param sender Sender address.
+/// @param owner Required sender address as an owner.
+error OwnerOnly(address sender, address owner);
+
+/// @dev Provided zero address.
+error ZeroAddress();
+
+/// @dev Guarded condition is detected.
+error Guarded();
+
+/// @title SafeGuard - Smart contract for Gnosis Safe Guard functionality
+/// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
+contract SafeGuard {
+    event OwnerUpdated(address indexed owner);
+    event ChangedGuard(address guard);
+    // keccak256("guard_manager.guard.address")
+    bytes32 internal constant GUARD_STORAGE_SLOT = 0x4a204f620c8c5ccdca3fd54d003badd85ba500436a431f0cbda4f558c93c34c8;
+
+    // Safe enum
+    enum Operation {Call, DelegateCall}
+
+    // Contract owner
+    address public owner;
+
+    /// @dev Guard constructor.
+    /// @param _owner Guard owner address.
+    constructor (address _owner) {
+        owner = _owner;
+    }
+
+    /// @dev Changes the owner address.
+    /// @param newOwner Address of a new owner.
+    function changeOwner(address newOwner) external virtual {
+        // Check for the ownership
+        if (msg.sender != owner) {
+            revert OwnerOnly(msg.sender, owner);
+        }
+
+        // Check for the zero address
+        if (newOwner == address(0)) {
+            revert ZeroAddress();
+        }
+
+        owner = newOwner;
+        emit OwnerUpdated(newOwner);
+    }
+
+    /// @dev Check transaction function implementation before the Safe's execute function call.
+    function checkTransaction(
+        address to,
+        uint256 value,
+        bytes memory data,
+        Operation operation,
+        uint256 safeTxGas,
+        uint256 baseGas,
+        uint256 gasPrice,
+        address gasToken,
+        address payable refundReceiver,
+        bytes memory signatures,
+        address msgSender
+    ) external {
+        // Right now we are blocking all the ETH funds transfer
+        if (value > 0) {
+            revert Guarded();
+        }
+    }
+
+    /// @dev Check transaction function implementation after the Safe's execute function call.
+    function checkAfterExecution(bytes32 txHash, bool success) external {
+
+    }
+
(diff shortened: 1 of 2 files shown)


contracts/multisigs/SafeGuard.sol
```diff
@@ -0,0 +1,88 @@
...
+/// @dev Guarded condition is detected.
+error Guarded();
+
+/// @title SafeGuard - Smart contract for Gnosis Safe Guard functionality
+/// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
+contract SafeGuard {
+    event OwnerUpdated(address indexed owner);
+    event ChangedGuard(address guard);
+    // keccak256("guard_manager.guard.address")
+    bytes32 internal constant GUARD_STORAGE_SLOT = 0x4a204f620c8c5ccdca3fd54d003badd85ba500436a431f0cbda4f558c93c34c8;
```
kupermind, line 21: This is the Safe guard slot that will be used in order to set a guard

contracts/multisigs/SafeGuard.sol
```diff
@@ -0,0 +1,88 @@
...
+        address gasToken,
+        address payable refundReceiver,
+        bytes memory signatures,
+        address msgSender
+    ) external {
+        // Right now we are blocking all the ETH funds transfer
+        if (value > 0) {
+            revert Guarded();
+        }
+    }
```
kupermind, line 70: The implementations of the `checkTransaction()` function that will be called by a guard before the Safe `execute()` function call

contracts/multisigs/SafeGuard.sol
```diff
@@ -0,0 +1,88 @@
...
+    /// @notice This function copies the corresponding Safe function such that it is correctly initialized.
+    /// @param guard The address of the guard to be used or the 0 address to disable the guard.
+    function setGuardForSafe(address guard) external {
+        bytes32 slot = GUARD_STORAGE_SLOT;
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            sstore(slot, guard)
+        }
+        emit ChangedGuard(guard);
+    }
```
kupermind, line 87: This function is called via a `delegatecall` from the Safe module setup when [Safe is created](https://github.com/safe-global/safe-contracts/blob/cb4b2b19b3e336b8defd3b8c9e0e6a2ae130598c/contracts/base/ModuleManager.sol#L32-L40) Since the delegatecall glues the storage space of the Safe instance being created and the SafeGuard contract, setting the guard slot would set it in the Safe instance.

test/GnosisSafeMultisig.js
```diff
@@ -132,5 +132,63 @@ describe("GnosisSafeMultisig", function () {
...
             expect(await multisig.getThreshold()).to.equal(threshold);
         });
+
+        it("Setting a guard contract address that guards transactions for not being able to transfer ETH funds", async function () {
+            // Set up a guard contract with the last Safe owner being the guard owner
+            const SafeGuard = await ethers.getContractFactory("SafeGuard");
+            const safeGuard = await SafeGuard.deploy(signers[2].address);
+            await safeGuard.deployed();
+
+            const to = safeGuard.address;
```
kupermind, line 142: Safe Guard address is the guard contract that is going to be called in the Safe modules setup, and is able to set the guard address slot value as it is called via a `delegatecall()` from the created Safe instance.

test/GnosisSafeMultisig.js
```diff
@@ -132,5 +132,63 @@ describe("GnosisSafeMultisig", function () {
...
+            const fallbackHandler = AddressZero;
+            const paymentToken = AddressZero;
+            const paymentReceiver = AddressZero;
+            const payment = 0;
+            let nonce = 0;
+            const payload = safeGuard.interface.encodeFunctionData("setGuardForSafe", [safeGuard.address]);
```
kupermind, line 148: The payload is included to be called by the modules setup, calling the Safe Guard contract with the specified function setting the guard address.

test/GnosisSafeMultisig.js
```diff
@@ -132,5 +132,63 @@ describe("GnosisSafeMultisig", function () {
...
+                expect(owners[i]).to.equal(defaultOwnerAddresses[i]);
+            }
+            expect(await multisig.getThreshold()).to.equal(threshold);
+
+            // Check the guard slot
+            const guardSlot = "0x4a204f620c8c5ccdca3fd54d003badd85ba500436a431f0cbda4f558c93c34c8";
+            let slotValue = await ethers.provider.getStorageAt(multisig.address, guardSlot);
+            slotValue = ethers.utils.hexlify(ethers.utils.stripZeros(slotValue));
+            const safeGuardAddress = safeGuard.address.toLowerCase();
+            expect(slotValue).to.equal(safeGuardAddress);
```
kupermind, line 176: Checking that the create Safe instance has the correctly set guard slot with the required address.

test/GnosisSafeMultisig.js
```diff
@@ -132,5 +132,63 @@ describe("GnosisSafeMultisig", function () {
...
+            const safeContracts = require("@gnosis.pm/safe-contracts");
+            nonce = await multisig.nonce();
+            // The function call is irrelevant, we just need to pass the value
+            let txHashData = await safeContracts.buildContractCall(multisig, "nonce", [], nonce, 0, 0);
+            txHashData.value = ethers.utils.parseEther("10");
+            const signMessageData = [await safeContracts.safeSignMessage(signers[1], multisig, txHashData, 0),
+                await safeContracts.safeSignMessage(signers[2], multisig, txHashData, 0)];
+            await expect(
+                safeContracts.executeTx(multisig, txHashData, signMessageData, 0)
+            ).to.be.revertedWith("Guarded");
```
kupermind, line 191: This set of actions simulates the Safe instance call execution that tries to transfer some value, and it is blocked by the Safe Guard.

77ph (approved):

PR #70, opened 2023-04-13
Setting multisig guards via Guard factory

kupermind:
- Proof of concept to set a multisig guar via the Guard factory;
- Preliminary test works, need to add more tests.

### DIFF
--- a/contracts/multisigs/SafeGuard.sol
+++ b/contracts/multisigs/SafeGuard.sol
@@ -1,6 +1,10 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.19;
 
+/// @dev The contract instance is already initialized.
+/// @param owner Contract instance owner address.
+error Initialized(address owner);
+
 /// @dev Only `owner` has a privilege, but the `sender` was provided.
 /// @param sender Sender address.
 /// @param owner Required sender address as an owner.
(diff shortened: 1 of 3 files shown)


contracts/multisigs/SafeGuard.sol
```diff
@@ -14,27 +18,32 @@ error Guarded();
...
-
     // Contract owner
     address public owner;
 
-    /// @dev Guard constructor.
-    /// @param _owner Guard owner address.
-    constructor (address _owner) {
+    /// @dev Initializes the Safe Guard instance by setting its owner.
+    /// @param _owner Safe Guard owner address.
+    function initialize (address _owner) external {
```
kupermind, line 32: Safe Guard contract is initializable via the Safe Guard factory now.

contracts/multisigs/SafeGuardFactory.sol
```diff
@@ -0,0 +1,144 @@
...
+        assembly {
+            guard := sload(slot)
+        }
+        if (guard != address(0)) {
+            revert SafeGuardALreadyExists(guard);
+        }
+
+        // Execute the payload, if provided
+        // The payload must be executed before the safe guard is set, as the payload could potentially break the guard
+        if (payloadLength > 0) {
```
kupermind, line 103: The custom payload is called also via a delegatecall(), as it would be called from the module manager of the GnosisSafe contract initialization.

contracts/multisigs/SafeGuardFactory.sol
```diff
@@ -0,0 +1,144 @@
...
+        // Supposedly the address(this) is the unique address of a safe contract that calls this guard creation function
+        // Supposedly the same Safe contract is not going to create a guard, delete it and try to create again
+        // in the same block. Thus, the combination of the Safe address and the block timestamp will be unique
+        bytes32 salt = keccak256(abi.encodePacked(address(this), block.timestamp));
+
+        // Create Safe Guard instance and write it into its designated slot
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            guard := create2(0, add(bytecode, 0x20), mload(bytecode), salt)
+        }
```
kupermind, line 129: Safe Guard instance is created, and the last multisig owner is set as the created Safe Guard owner.

contracts/test/DelegateCall.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.19;
+
+contract DelegateCall {
```
kupermind, line 4: Test DelegateCall contract

PR #74, opened 2023-04-24
Continue the implementation of the service registry with the custom ERC20 token security

kupermind:
- Added `update()` function in the new `ServiceManager` contract;
- Adding tests.

### DIFF
--- a/contracts/ServiceManagerToken.sol
+++ b/contracts/ServiceManagerToken.sol
@@ -60,6 +60,7 @@ contract ServiceManagerToken is GenericManager {
     /// @param agentIds Canonical agent Ids.
     /// @param agentParams Number of agent instances and required bond to register an instance in the service.
     /// @param threshold Threshold for a multisig composed by agents.
+    /// @return serviceId Created service Id.
     function create(
         address serviceOwner,
         address token,
@@ -67,7 +68,7 @@ contract ServiceManagerToken is GenericManager {
         uint32[] memory agentIds,
         IService.AgentParams[] memory agentParams,
         uint32 threshold
-    ) external returns (uint256)
+    ) external returns (uint256 serviceId)
     {
         // Check if the minting is paused
         if (paused) {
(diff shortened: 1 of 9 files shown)


77ph:
Everything found is in internal audit.

77ph (approved):

contracts/ServiceManagerToken.sol
```diff
@@ -32,14 +32,22 @@ contract ServiceManagerToken is GenericManager {
     // Operator whitelist address
     address public operatorWhitelist;
 
+    /// @dev ServiceRegistryTokenUtility constructor.
+    /// @param _serviceRegistry Service Registry contract address.
+    /// @param _serviceRegistryTokenUtility Service Registry Token Utility contract address.
     constructor(address _serviceRegistry, address _serviceRegistryTokenUtility) {
```
mariapiamo, line 38: Ok, here we will also add the operator whitelist address and the corresponding non-zero check

contracts/ServiceManagerToken.sol
```diff
@@ -107,22 +114,56 @@ contract ServiceManagerToken is GenericManager {
...
+            revert ZeroAddress();
+        }
+
+        if (token == ETH_TOKEN_ADDRESS) {
+            // Call the original ServiceRegistry contract function
+            IService(serviceRegistry).update(msg.sender, configHash, agentIds, agentParams, threshold, serviceId);
+            // Reset the service token-based data
+            // This function still needs to be called as the previous token could be a custom ERC20 token
+            IServiceTokenUtility(serviceRegistryTokenUtility).resetServiceToken(serviceId);
+        } else {
```
mariapiamo, line 144: The following will be fixed

contracts/ServiceRegistryTokenUtility.sol
```diff
@@ -148,28 +155,35 @@ contract ServiceRegistryTokenUtility is IErrorsRegistries {
     }
 
     /// @dev Creates a record with the token-related information for the specified service.
+    /// @notice We assume that the token is checked for being a non-zero address and a non-ETH address representation
+    ///         outside of this function. Here we optimistically check for the token to have a specific `balanceOf`
```
mariapiamo, line 159: what do you mean by "we optimistically check"?

mariapiamo: All the notice is not super clear to me, maybe can be rephrased a bit

contracts/ServiceRegistryTokenUtility.sol
```diff
@@ -148,28 +155,35 @@ contract ServiceRegistryTokenUtility is IErrorsRegistries {
...
-        if (token == address(0)) {
-            revert ZeroAddress();
-        }
         // TODO Check for the token ERC20 formality
         
         uint256 securityDeposit;
         // Service is newly created and all the array lengths are checked by the original ServiceRegistry create() function
         for (uint256 i = 0; i < agentIds.length; ++i) {
-            // Check for a non-zero bond value
+            // Check for a non-zero bond value and skip those with zeros (possible when updating a service)
```
mariapiamo, line 184: when this function is used to `update` the service, we can have some bonds equal to zero. But we still need to check that there is at least one i such that bond[i] is not zero, correct?

mariapiamo: Ah no, I see from the next PR that a check is made at the serviceManager level. Correct?

contracts/ServiceManagerToken.sol
```diff
@@ -107,22 +114,56 @@ contract ServiceManagerToken is GenericManager {
...
     ) external returns (bool)
     {
-        return IService(serviceRegistry).update(msg.sender, configHash, agentIds, agentParams,
-            threshold, serviceId);
+        // Check for the zero address
+        if (token == address(0)) {
+            revert ZeroAddress();
+        }
+
+        if (token == ETH_TOKEN_ADDRESS) {
```
mariapiamo, line 138: here we will add the zero-bond check

contracts/ServiceManagerToken.sol
```diff
@@ -107,22 +114,56 @@ contract ServiceManagerToken is GenericManager {
...
+            // This function still needs to be called as the previous token could be a custom ERC20 token
+            IServiceTokenUtility(serviceRegistryTokenUtility).resetServiceToken(serviceId);
+        } else {
+            // Wrap agent params with just 1 WEI bond going to the original ServiceRegistry contract,
+            // and actual token bonds being recorded with the ServiceRegistryTokenUtility contract
+            uint256 numAgents = agentParams.length;
+            uint256[] memory bonds = new uint256[](numAgents);
+            for (uint256 i = 0; i < numAgents; ++i) {
+                // Copy actual bond values for each agent Id that has at least one slot in the updated service
+                if (agentParams[i].slots > 0) {
```
mariapiamo, line 151: Now, if agentParams[i].slots > 0, we also need to make sure that agentParams[i].bonds> 0

contracts/ServiceRegistryTokenUtility.sol
mariapiamo: clarified in the next pr

contracts/ServiceRegistryTokenUtility.sol
mariapiamo: yes

contracts/ServiceManagerToken.sol
mariapiamo: added in the next PR

contracts/ServiceManagerToken.sol
mariapiamo: added in the next PR

PR #76, opened 2023-04-26
refactor and test: account for internal audit findings

kupermind:
- Account for the audit findings;
- Coverage is back to 100%.


contracts/ServiceRegistryTokenUtility.sol
```diff
@@ -254,15 +373,19 @@ contract ServiceRegistryTokenUtility is IErrorsRegistries {
...
-            // TODO Re-entrancy
-            // TODO Safe transferFrom
-            bool success = IToken(token).transferFrom(serviceOwner, address(this), securityDeposit);
-            if (!success) {
-                revert TransferFailed(token, serviceOwner, address(this), securityDeposit);
+            uint256 balanceBefore = IToken(token).balanceOf(address(this));
+            safeTransferFrom(token, serviceOwner, address(this), securityDeposit);
+            uint256 balanceAfter = IToken(token).balanceOf(address(this));
+            // Check the correctness of received funds
+            if (balanceBefore > balanceAfter || (balanceAfter - balanceBefore) != securityDeposit) {
```
mariapiamo, line 384: bigger or equal?

77ph: bigger for avoid x - y < 0

contracts/ServiceManagerToken.sol
```diff
@@ -31,32 +31,32 @@ contract ServiceManagerToken is GenericManager {
...
+    // Service manager version number
+    string public constant VERSION = "1.1.0";
 
     /// @dev ServiceRegistryTokenUtility constructor.
     /// @param _serviceRegistry Service Registry contract address.
     /// @param _serviceRegistryTokenUtility Service Registry Token Utility contract address.
-    constructor(address _serviceRegistry, address _serviceRegistryTokenUtility) {
-        // Check for the zero address
+    constructor(address _serviceRegistry, address _serviceRegistryTokenUtility, address _operatorWhitelist) {
+        // Check for the Service Registry related contract zero addresses
```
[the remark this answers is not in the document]
kupermind: Since we can skip it here, and in general delete it by setting to `address(0)`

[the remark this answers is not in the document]
77ph: possible combination: as disable operatorWhitelist  by default in deploy

contracts/ServiceRegistryTokenUtility.sol
kupermind: First check is the reverse one - if before is bigger than after, meaning that during the transfer the contract has lost funds :) as if the token is really bad. In this case we can't get the diff, safe math would fail going underflow.

contracts/OperatorWhitelist.sol
```diff
@@ -48,24 +50,31 @@ contract OperatorWhitelist {
...
     /// @param operators Set of operator addresses.
     /// @param statuses Set of whitelisting statuses.
-    function setOperatorsStatuses(uint256 serviceId, address[] memory operators, bool[] memory statuses) external {
+    /// @param setCheck True if the whitelisting check is needed, and false otherwise.
+    function setOperatorsStatuses(
+        uint256 serviceId,
+        address[] memory operators,
+        bool[] memory statuses,
+        bool setCheck
+    ) external {
```
kupermind, line 77: Extended the list of parameters as per specs

contracts/ServiceManagerToken.sol
```diff
@@ -31,32 +31,32 @@ contract ServiceManagerToken is GenericManager {
...
         // Check for the contract ownership
         if (msg.sender != owner) {
             revert OwnerOnly(msg.sender, owner);
         }
 
-        // Check for the zero address
-        if (newOperatorWhitelist == address(0)) {
-            revert ZeroAddress();
-        }
-
```
kupermind, line 59: We can cancel the whitelister by setting its address to zero

contracts/ServiceManagerToken.sol
```diff
@@ -98,6 +98,11 @@ contract ServiceManagerToken is GenericManager {
             uint256 numAgents = agentParams.length;
             uint256[] memory bonds = new uint256[](numAgents);
             for (uint256 i = 0; i < numAgents; ++i) {
+                // Check for the zero bond value
+                if (agentParams[i].bond == 0) {
+                    revert ZeroValue();
+                }
```
kupermind, line 104: Bonds can never be zero as this way we would wrap zero values on the original ServiceRegistry side

contracts/ServiceManagerToken.sol
```diff
@@ -149,21 +154,24 @@ contract ServiceManagerToken is GenericManager {
...
+                    // Check for the zero bond value
+                    if (agentParams[i].bond == 0) {
+                        revert ZeroValue();
+                    }
                     bonds[i] = agentParams[i].bond;
+                    // Wrap bonds with the BOND_WRAPPER value for the original ServiceRegistry contract
+                    agentParams[i].bond = BOND_WRAPPER;
                 }
-                // Wrap bonds with the BOND_WRAPPER value for the original ServiceRegistry contract
-                agentParams[i].bond = BOND_WRAPPER;
```
kupermind, line 155: Fixed a bug and check for the bond value to be bigger than zero when slots are bigger than zero as well

contracts/ServiceRegistryTokenUtility.sol
```diff
@@ -322,17 +445,21 @@ contract ServiceRegistryTokenUtility is IErrorsRegistries {
...
-
-            // Transfer tokens from the operator account
-            // TODO Re-entrancy
-            // TODO Safe transferFrom
-            bool success = IToken(token).transferFrom(operator, address(this), totalBond);
-            if (!success) {
-                revert TransferFailed(token, operator, address(this), totalBond);
-            }
+            // Update operator's bonding balance
+            mapOperatorAndServiceIdOperatorBalances[operatorService] += totalBond;
```
kupermind, line 449: Fixed a bug with transferring more than the current totalBond

contracts/test/ReentrancyTokenAttacker.sol
```diff
@@ -0,0 +1,103 @@
...
+    /// @return refund Returned bonded token amount, or zero if the service is ETH-secured.
+    function unbondTokenRefund(address operator, uint256 serviceId) external returns (uint256 refund);
+
+    /// @dev Drains slashed funds to the drainer address.
+    /// @param token Token address.
+    /// @return amount Drained amount.
+    function drain(address token) external returns (uint256 amount);
+}
+
+contract ReentrancyTokenAttacker {
```
kupermind, line 39: Token reentrancy attacker to test all the possible token transfer failures and reentrancy

77ph (approved):

contracts/ServiceRegistryTokenUtility.sol
mariapiamo: Yeah, instead equal to zero is handled by the second operator

contracts/ServiceManagerToken.sol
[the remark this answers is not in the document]
mariapiamo: Yeah, but since we already give the possibility to SetCheck equal to false, I do not see the real motivation to further allow the DAO to disable the operatorWhitelist in this way. A governance vote would still be required to disable/enable it and in the eventuality, DAO wants to completely disable the operatorWhitelist a new manager contract can be considered.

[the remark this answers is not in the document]
kupermind: Disabling is still much cheaper than deploying a new contract

README.md
```diff
@@ -37,9 +37,12 @@ A more detailed set of registries definitions are provided [here](https://github
   - [ComponentRegistry](https://github.com/valory-xyz/autonolas-registries/blob/main/contracts/ComponentRegistry.sol)
   - ServiceRegistry [L1](https://github.com/valory-xyz/autonolas-registries/blob/main/contracts/ServiceRegistry.sol),
     [L2](https://github.com/valory-xyz/autonolas-registries/blob/main/contracts/ServiceRegistry2.sol)
```
mariapiamo, line 39:
```suggestion
    [L2](https://github.com/valory-xyz/autonolas-registries/blob/main/contracts/ServiceRegistryL2.sol)
```

mariapiamo: broken link

contracts/ServiceManagerToken.sol
```diff
@@ -137,7 +142,7 @@ contract ServiceManagerToken is GenericManager {
 
         if (token == ETH_TOKEN_ADDRESS) {
```
mariapiamo, line 143: here we will add the zero-bond check

mariapiamo (approved):

PR #100, opened 2023-06-07
feat: first solana integration try

kupermind:
- Initial solana integration.


77ph (approved):

kupermind:
This initial development was still a very distant version compared to where it evolved to.

PR #107, opened 2023-08-08
feat: Conditional Tokens lib

kupermind:
- Adding the possibility to correctly call the Conditional Tokens contract.


kupermind:
Migrated to:
https://github.com/valory-xyz/contract-utils

PR #116, opened 2023-09-21
feat: first implementation of the service staking

kupermind:
- First implementation of the service staking.


contracts/staking/ServiceStakingToken.sol
```diff
@@ -0,0 +1,199 @@
...
+}
+
+/// @dev Failure of a transfer.
+/// @param token Address of a token.
+/// @param from Address `from`.
+/// @param to Address `to`.
+/// @param value Value.
+error TransferFailed(address token, address from, address to, uint256 value);
+
+/// @title ServiceStakingToken - Smart contract for staking the service by its owner having an ERC20 token as a deposit
```
DavidMinarsch, line 30:
```suggestion
/// @title ServiceStakingToken - Smart contract for staking a service by its owner when the service has an ERC20 token as the deposit
```

contracts/staking/ServiceStakingToken.sol
```diff
@@ -0,0 +1,199 @@
...
+    // Security token address for staking corresponding to the service deposit token
+    address public immutable securityToken;
+
+    /// @dev ServiceStakingToken constructor.
+    /// @param _apy Staking APY (in single digits).
+    /// @param _minSecurityDeposit Minimum security deposit for a service to be eligible to stake.
+    /// @param _stakingRatio Staking ratio: number of seconds per nonce (in 18 digits).
+    /// @param _serviceRegistry ServiceRegistry contract address.
+    /// @param _serviceRegistryTokenUtility ServiceRegistryTokenUtility contract address.
+    /// @param _securityToken Address of a service security token.
```
DavidMinarsch, line 46: _securityToken is not a suitable name. Let's call it _stakingToken?

contracts/staking/ServiceStakingToken.sol
```diff
@@ -0,0 +1,199 @@
...
+    /// @notice The implementation is fully copied from the audited MIT-licensed solmate code repository:
+    ///         https://github.com/transmissions11/solmate/blob/v7/src/utils/SafeTransferLib.sol
+    ///         The original library imports the `ERC20` abstract token contract, and thus embeds all that contract
+    ///         related code that is not needed. In this version, `ERC20` is swapped with the `address` representation.
+    ///         Also, the final `require` statement is modified with this contract own `revert` statement.
+    /// @param token Token address.
+    /// @param from Address to transfer tokens from.
+    /// @param to Address to transfer tokens to.
+    /// @param amount Token amount.
+    function safeTransferFrom(address token, address from, address to, uint256 amount) internal {
```
DavidMinarsch, line 77: Why does this need to be copied here?

contracts/staking/ServiceStakingToken.sol
```diff
@@ -0,0 +1,199 @@
...
+    /// @dev Safe token transfer implementation.
+    /// @notice The implementation is fully copied from the audited MIT-licensed solmate code repository:
+    ///         https://github.com/transmissions11/solmate/blob/v7/src/utils/SafeTransferLib.sol
+    ///         The original library imports the `ERC20` abstract token contract, and thus embeds all that contract
+    ///         related code that is not needed. In this version, `ERC20` is swapped with the `address` representation.
+    ///         Also, the final `require` statement is modified with this contract own `revert` statement.
+    /// @param token Token address.
+    /// @param to Address to transfer tokens to.
+    /// @param amount Token amount.
+    function safeTransfer(address token, address to, uint256 amount) internal {
```
DavidMinarsch, line 119: Same question

contracts/staking/ServiceStakingToken.sol
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    /// @dev Deposits funds for staking.
+    /// @param amount Token amount to deposit.
+    function deposit(uint256 amount) external {
+        // Distribute current staking rewards
+        _checkpoint(0);
+
+        // Add to the overall balance
+        safeTransferFrom(securityToken, msg.sender, address(this), amount);
```
DavidMinarsch, line 185: Why can we not work with an interface approach? This way of doing it looks messy to me.

DavidMinarsch: We should assume an IERC20 interface https://ethereum.org/en/developers/docs/standards/tokens/erc-20/ and so the "safe" versions are not present. Will have to add appropriate checks accordingly

contracts/staking/ServiceStakingToken.sol
```diff
@@ -0,0 +1,199 @@
...
+        safeTransfer(securityToken, to, amount);
+
+        emit Withdraw(to, amount);
+    }
+
+    /// @dev Deposits funds for staking.
+    /// @param amount Token amount to deposit.
+    function deposit(uint256 amount) external {
+        // Distribute current staking rewards
+        _checkpoint(0);
```
DavidMinarsch, line 182: I am not sure it's sensible to do this call. Is it a must or did you add it for "convenience". In the later case I'd remove it.

contracts/staking/ServiceStakingToken.sol
```diff
@@ -0,0 +1,199 @@
...
+
+        // Add to the overall balance
+        safeTransferFrom(securityToken, msg.sender, address(this), amount);
+
+        // Add to the overall balance
+        uint256 newBalance = balance + amount;
+
+        // Update rewards per second
+        uint256 newRewardsPerSecond = (newBalance * apy) / (100 * 365 days);
```
DavidMinarsch, line 191: This strikes me as odd.

DavidMinarsch: Don't like how coupled transferring balance in is with checkpointing logic.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,278 @@
...
+
+// Multisig interface
+interface IMultisig {
+    /// @dev Gets the multisig nonce.
+    /// @return Multisig nonce.
+    function nonce() external returns (uint256);
+}
+
+// Service Registry interface
+interface IService {
```
DavidMinarsch, line 12: Should have their own files imo?

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,278 @@
...
+    function _checkpoint(uint256 serviceId) internal returns (uint256 idx) {
+        // Get the service Id set length
+        uint256 size = setServiceIds.length;
+        uint256 lastBalance = balance;
+
+        // If the balance is zero, just bump the timestamp for each of the staking service
+        if (lastBalance == 0) {
+            for (uint256 i = 0; i < size; ++i) {
+                // Set the current timestamp
+                mapServiceInfo[setServiceIds[i]].ts = block.timestamp;
```
DavidMinarsch, line 160: I don't like this approach. It will be very hard to see when a service started staking (one needs to look at events and reconstruct). I'd prefer we have two fields in `mapServiceInfo`, one for `start` and one for `lastCheckpoint`

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,278 @@
...
+                }
+
+                // Add the calculated reward to the service info
+                curInfo.reward += reward;
+
+                // Adjust the starting ts for each service to a current timestamp
+                curInfo.ts = block.timestamp;
+            }
+            // Update the storage balance
+            balance = lastBalance;
```
DavidMinarsch, line 231: We should probably have two balance variables: `balance` should simply show the total balance currently in the contract. Then `availableRewards` is the balance that's still available to stakers. Again, good to have this separate for frontends and to easily inspect

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,278 @@
...
+    /// @param serviceId Service Id.
+    function unstake(uint256 serviceId) external {
+        ServiceInfo storage sInfo = mapServiceInfo[serviceId];
+        // Check for the service ownership
+        if (msg.sender != sInfo.owner) {
+            revert();
+        }
+
+        // Call the checkpoint and get the service index in the set of services
+        uint256 idx = _checkpoint(serviceId);
```
DavidMinarsch, line 250: We don't need this call. We simply need to update the two balances and then checkpoint is only called when people unstake. Keeps it simpler.

Also, I think we should have checkpoint being public so anyone can also just call it to update all reward allocations to services.

Also, I think there should be a getter method that virtualises the "would be" if one called checkpoint. Then the frontend can easily show current staking rewards for staked services.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,278 @@
...
+        uint256 amount = sInfo.reward;
+        uint256 lastBalance = balance;
+        if (lastBalance < minBalance) {
+            amount += lastBalance;
+            balance = 0;
+        }
+
+        // Transfer accumulated rewards to the service owner
+        if (amount > 0) {
+            _withdraw(msg.sender, amount);
```
DavidMinarsch, line 265: Needs to be sent to multisig, not service owner!

contracts/staking/ServiceStakingToken.sol
kupermind: It is provided as a library in solmate, however it pulls the full ERC20 token contract for that, and we don't want that.

Just figured it's not needed at all as our secured tokens are already checked for "safe" when depositing the first time in the ServiceRegistryTokenUtility contract.

contracts/staking/ServiceStakingToken.sol
kupermind: See above

contracts/staking/ServiceStakingToken.sol
kupermind: See above

contracts/staking/ServiceStakingToken.sol
kupermind: I can't see how else we can account correct rewards for all the services with the current balance that is going to change. We need to set a checkpoint for this current time, then increase the balance and calculate future rewards based on a new balance and starting with a new time.

contracts/staking/ServiceStakingToken.sol
77ph: A more detailed history of why it appeared "safe" ERC20:
https://medium.com/coinmonks/missing-return-value-bug-at-least-130-tokens-affected-d67bf08521ca

In fact, it has been discussed for a long time to make this part of the standard:
https://eips.ethereum.org/EIPS/eip-4524:
https://ethereum-magicians.org/t/why-isnt-there-an-erc-for-safetransfer-for-erc-20/7604/21

contracts/staking/ServiceStakingBase.sol
kupermind: Agree with last two, but not the first one. First, if we want to maintain the set of staked services, we need to do the correct cleaning. Second, we can't just update balances without doing a checkpoint, as it would break the math of when different balances were added, see above.

contracts/staking/ServiceStakingToken.sol
kupermind: The balance is added, the reward per second changes. I don't see the other way of doing that. Still think we can't mix different balance deposit over different times, as in my example above.

Bottom line - we need a checkpoint on deposits and on unstakes.

contracts/staking/ServiceStakingToken.sol
kupermind: We can extract it as a standalone library in the `utils` folder

77ph: Ref:
https://github.com/valory-xyz/autonolas-registries/blob/main/contracts/ServiceRegistryTokenUtility.sol#L383
That is, our assumptions about the token must coincide with those we made earlier.

contracts/staking/ServiceStakingBase.sol
77ph: I understand correctly that receiving a reward is not a mirror (asymmetrical) operation to "staking"

contracts/staking/ServiceStakingBase.sol
kupermind: Will be used when possible, as we don't want the bytecode of our original contracts changed if original interfaces are updated. Now I remember that was the reason not to add to existing files.

contracts/staking/ServiceStakingToken.sol
DavidMinarsch: Why not? You can simply first update the balances? Anything that happened at t you can do at t+k.

contracts/staking/ServiceStakingBase.sol
DavidMinarsch: Yes the service logic is responsible for deciding where rewards go.

contracts/staking/ServiceStakingToken.sol
kupermind: The balances can be updated only when calculating rewards for all the eligible services. To do this, need to traverse all of them and sync at the current timestamp.

kupermind: Otherwise we'll have to keep track of all the balance deposits and their times.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,248 @@
...
+    uint256 public minBalance;
+    // Rewards per second
+    uint256 public rewardsPerSecond;
+    // Mapping of serviceId => staking service info
+    mapping (uint256 => ServiceInfo) public mapServiceInfo;
+    // Set of currently staking serviceIds
+    uint256[] public setServiceIds;
+
+    /// @dev ServiceStakingBase constructor.
+    /// @param _apy Staking APY (in single digits).
```
DavidMinarsch, line 97: apy - not sure that's appropriate terminology

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,248 @@
...
+    /// @param _minStakingDeposit Minimum security deposit for a service to be eligible to stake.
+    /// @param _stakingRatio Staking ratio: number of seconds per nonce (in 18 digits).
```
DavidMinarsch, line 99:
```suggestion
    /// @param _livenessRatio Liveness ratio: number of seconds per nonce (in 18 digits).
```
seems more appropriate

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,248 @@
...
+                // Calculate the staking nonce ratio
+                uint256 curNonce = IMultisig(curInfo.multisig).nonce();
+                // Get the last service checkpoint: staking start time or the global checkpoint timestamp
+                uint256 serviceCheckpoint = tsCheckpointLast;
+                // Adjust the service checkpoint time if the service was staking less than the current staking period
+                if (curInfo.tsStart > tsCheckpointLast) {
+                    serviceCheckpoint = curInfo.tsStart;
+                }
+                // Calculate the nonce ratio in 1e18 value
+                uint256 ratio = ((block.timestamp - serviceCheckpoint) * 1e18) / (curNonce - curInfo.nonce);
```
DavidMinarsch, line 189: this is the wrong way round. More tx is more liveness!

DavidMinarsch: Also prevent division by zero

DavidMinarsch: I can see an issue here with closeness of checkpoints leading to rounding issues

contracts/staking/ServiceStakingToken.sol
```diff
@@ -0,0 +1,99 @@
...
+
+        // Add to the overall balance
+        SafeTransferLib.safeTransferFrom(stakingToken, msg.sender, address(this), amount);
+
+        // Add to the contract and available rewards balances
+        uint256 newBalance = balance + amount;
+        uint256 newAvailableRewards = availableRewards + amount;
+
+        // Update rewards per second
+        uint256 newRewardsPerSecond = (newAvailableRewards * apy) / (100 * 365 days);
```
DavidMinarsch, line 123: I'd prefer if rewards per second was an arg rather than apy

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,248 @@
...
+                // it is equal to at most the tsCheckpoint of the last deposit happening during every _checkpoint() call
+                uint256 reward = rewardsPerSecond * (block.timestamp - serviceCheckpoint);
+                // Adjust the reward if it goes out of calculated max bounds
+                if (reward > maxRewardsPerService) {
+                    reward = maxRewardsPerService;
+                }
+
+                // Adjust the available rewards value
+                if (lastAvailableRewards >= reward) {
+                    lastAvailableRewards -= reward;
```
DavidMinarsch, line 232: use memory var in loop when optimising later

contracts/staking/ServiceStakingToken.sol
DavidMinarsch: ok

contracts/staking/ServiceStakingToken.sol
DavidMinarsch: we are talking about "balance" and "availableRewards", both of which can be updated without problem imo. Just don't see where in the code the issue is

contracts/staking/ServiceStakingToken.sol
DavidMinarsch: no reward per second should be a constant
I see now what you misunderstood from the spec.

Nonces should purely be used for liveness check! It's a binary condition. Are enough tx made in a given time window? If yes, you get rewards as per "rewardsPerSecond" specifier. If not, no rewards.

contracts/staking/ServiceStakingBase.sol
DavidMinarsch: True

contracts/staking/ServiceStakingToken.sol
kupermind: I believe this is exactly what's implemented `stakingRatio` or the `livenessCheck` is the amount of multisig tx-s made in a specific timeframe. If it is above the defined value, the service is eligible for reward, otherwise it's not. Rewards do not depend on the ratio, just a yes or no.

The available reward balance is then split between eligible services (satisfying the liveness check) from the time of the last checkpoint until a current block timestamp.

contracts/staking/ServiceStakingBase.sol
kupermind: This can be defined in either direction. If we believe there will be more tx-s than seconds, then this must be reversed. We are still checking on a specific period of time (from last checkpoint till now) and how many tx-s were performed by every staking service within that period.

77ph: Agree, incorrect.
It should be tps with dimension: nonce/s (tx/s). In the code we got the inverse value.

contracts/staking/ServiceStakingToken.sol
77ph: So as I understand what we want:
A given constant RewardsPerSecond (RPS) with dimension [T/s] T - token (including ETH)
RewardsPerSecond : immutable, and accordingly does not depend on the current AvailableRewards
Formula reward =  RPS * (t2 - t1) (1*)

A simple example of how this works without boundary conditions ( AvailableRewards >> sum(rewards)):
0. Start staking with AvailableRewards = 1M Tokens, no staking services, RPS = 1T/s
1. Staking 3 services: s1, s2, s3
2. +24h. Unstaking 3 services
s1 = pass tps_ratio => reward1 (yes)
s2 = pass tps_ratio => reward2 (yes)
s3 = fail tps_ratio => 0 (no)
reward1 = RPS * 24*60*60s = 1T/s * 86400s = 86400 T
reward2 = RPS * 24*60*60s = 1T/s * 86400s = 86400 T
AvailableRewards =  AvailableRewards - sum(rewards) = 1M - 2*86400 = 827,200

Let's assume we still have  AvailableRewards = 100k
0. No staking atm
1. Staking 3 services: s1, s2, s3
2. +24h. Unstaking 3 services
s1 = pass tps_ratio => reward1 (yes)
s2 = pass tps_ratio => reward2 (yes)
s3 = fail tps_ratio => 0 (no)
reward1 = RPS * 24*60*60s = 1T/s * 86400s = 86400 T (incorrect)
reward2 = RPS * 24*60*60s = 1T/s * 86400s = 86400 T (incorrect)
AvailableRewards =  AvailableRewards - sum(rewards) = 100k - 2*86400 < 0 
If AvailableRewards < sum(rewards) by (1*):
We need to distribute rewards based on the current limit in proportion to the "ideal" reward (1*)
reward1fix = RPS * 24*60*60s /sum(rewards) * AvailableRewards  = 1/2 * 100k = 50k T
reward2fix = RPS * 24*60*60s /sum(rewards) * AvailableRewards  = 1/2 * 100k = 50k T
AvailableRewards = AvailableRewards - 100k = 0

mariapiamo: With your suggestion @DavidMinarsch and the formula of @77ph the contract becomes significantly simpler. However, I have concerns because staking rewards are entirely unrelated to the service security deposit amount. This could potentially lead to a situation where staking rewards greatly exceed the service security deposits.

DavidMinarsch: "I have concerns because staking rewards are entirely unrelated to the service security deposit amount." -> that's up to the contract deployer. They set both after all.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,296 @@
...
+    /// @param amount Amount to withdraw.
+    function _withdraw(address to, uint256 amount) internal virtual {}
+
+    /// @dev Stakes the service.
+    /// @param serviceId Service Id.
+    function stake(uint256 serviceId) external {
+        // Check the service conditions for staking
+        (uint96 stakingDeposit, address multisig, , , , , uint8 state) = IService(serviceRegistry).mapServices(serviceId);
+        // The service must be deployed
+        if (state != 4) {
```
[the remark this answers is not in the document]
DavidMinarsch: That's the point yes

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,309 @@
...
+                // Get the last service checkpoint: staking start time or the global checkpoint timestamp
+                uint256 serviceCheckpoint = tsCheckpointLast;
+                // Adjust the service checkpoint time if the service was staking less than the current staking period
+                if (curInfo.tsStart > tsCheckpointLast) {
+                    serviceCheckpoint = curInfo.tsStart;
+                }
+                // Calculate the liveness ratio in 1e18 value
+                uint256 ratio;
+                // If the checkpoint was called in the exactly same block, the ratio is zero
+                if (block.timestamp > tsCheckpointLast) {
```
DavidMinarsch, line 194: for readability
```suggestion
                if (block.timestamp > serviceCheckpoint) {
```

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,309 @@
...
+
+        // Add the service Id to the set of staked services
+        setServiceIds.push(serviceId);
+
+        emit ServiceStaked(serviceId, msg.sender);
+    }
+
+    /// @dev Checkpoint to allocate rewards up until a current time.
+    /// @param serviceId Service Id that unstakes, or 0 if the function is called during the deposit of new funds.
+    function _checkpoint(uint256 serviceId) internal returns (uint256 idx) {
```
DavidMinarsch, line 157: Don't see why _checkpoint has to depend on a serviceId at all.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,309 @@
...
+                    ++numServices;
+                }
+
+                // Record current service multisig nonce
+                serviceNonces[i] = curNonce;
+
+                // Record the unstaked service Id index in the global set of staked service Ids
+                if (serviceIds[i] == serviceId) {
+                    idx = i;
+                }
```
DavidMinarsch, line 214: Should not happen in this scope imo. You have a tendency to combine a lot of individual responsibilities

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,309 @@
...
+    /// @param serviceId Service Id.
+    function unstake(uint256 serviceId) external {
+        ServiceInfo storage sInfo = mapServiceInfo[serviceId];
+        // Check for the service ownership
+        if (msg.sender != sInfo.owner) {
+            revert OwnerOnly(msg.sender, sInfo.owner);
+        }
+
+        // Call the checkpoint and get the service index in the set of services
+        uint256 idx = _checkpoint(serviceId);
```
DavidMinarsch, line 281: I think there should be a simple call to `_checkpoint()` and any service specific stuff to serviceID should happen outside this context. That keeps the code simpler to reason about.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,296 @@
...
+/// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
+/// @author Andrey Lebedev - <andrey.lebedev@valory.xyz>
+/// @author Mariapia Moscatiello - <mariapia.moscatiello@valory.xyz>
+abstract contract ServiceStakingBase is IErrorsRegistries {
+    event ServiceStaked(uint256 indexed serviceId, address indexed owner);
+    event Checkpoint(uint256 indexed balance);
+    event ServiceUnstaked(uint256 indexed serviceId, address indexed owner, uint256 reward);
+    event Deposit(address indexed sender, uint256 amount, uint256 newBalance, uint256 newAvailableRewards,
+        uint256 rewardsPerSecond);
+    event Withdraw(address indexed to, uint256 amount);
```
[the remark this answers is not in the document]
kupermind: good point, however its place is in the `ServiceUnstaked` event

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        apy = _apy;
+        minStakingDeposit = _minStakingDeposit;
+        stakingRatio = _stakingRatio;
+        serviceRegistry = _serviceRegistry;
+    }
+
+    /// @dev Checks token / ETH staking deposit.
+    /// @param stakingDeposit Staking deposit.
```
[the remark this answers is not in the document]
kupermind: No, it's not needed here as this param is not used.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,296 @@
...
+
+                // Record the unstaked service Id index in the global set of staked service Ids
+                if (curServiceId == serviceId) {
+                    idx = i;
+                }
+            }
+
+            // Process each eligible service Id reward
+            // Calculate the maximum possible reward per service during the last deposit period
+            uint256 maxRewardsPerService = (rewardsPerSecond * (block.timestamp - tsCheckpointLast)) / numServices;
```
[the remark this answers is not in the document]
kupermind: This is not applicable anymore

contracts/staking/ServiceStakingBase.sol
kupermind: To get the index of a service being unstaked. I don't want it there myself and I keep thinking on how to exclude it, if possible. But gas optimization is the priority.

contracts/staking/ServiceStakingBase.sol
kupermind: Yeah, that's why I'm usually gas efficient ;)

contracts/staking/ServiceStakingBase.sol
kupermind: As stated above, if there's a way not to spend more gas on figuring out the service Id index, I'd do it.

kupermind: It's several times per day the trade off between `pretty` and `gas efficient`.

kupermind: ok I've refactored it, but still not super happy, will see in the morning

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,384 @@
...
+                    // Calculate the updated reward
+                    uint256 updatedReward = (eligibleServiceRewards[i] * lastAvailableRewards) / totalRewards;
+                    // Add to the total updated reward
+                    totalRewards += updatedReward;
+                    // Add reward to the service overall reward
+                    uint256 curServiceId = eligibleServiceIds[i];
+                    mapServiceInfo[curServiceId].reward += updatedReward;
+                }
+
+                // If the reward adjustment happened to have small leftovers, add it to the last traversed service
```
mariapiamo, line 263: So last service in the list can receive a bit more?

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,384 @@
...
+        ServiceInfo storage sInfo = mapServiceInfo[serviceId];
+        sInfo.multisig = multisig;
+        sInfo.owner = msg.sender;
+        sInfo.nonce = IMultisig(multisig).nonce();
+        sInfo.tsStart = block.timestamp;
+
+        // Add the service Id to the set of staked services
+        setServiceIds.push(serviceId);
+
+        emit ServiceStaked(serviceId, msg.sender);
```
mariapiamo, line 156: Can we also emit the service multisig address and the nonce at the time of the stake?

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,384 @@
...
+            }
+        }
+
+        // Transfer the service back to the owner
+        IService(serviceRegistry).transferFrom(address(this), msg.sender, serviceId);
+
+        // Send the remaining small balance along with the reward if it is below the chosen threshold
+        uint256 amount = sInfo.reward;
+        uint256 lastAvailableRewards = availableRewards;
+        if (lastAvailableRewards < minBalance) {
```
mariapiamo, line 327: So if into the contract lastAvailableRewards is smaller than minBalance, the first service that is going to call the unstake will receive it

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,384 @@
...
+
+        // Clear all the data about the unstaked service
+        // Delete the service info struct
+        delete mapServiceInfo[serviceId];
+
+        // Update the set of staked service Ids
+        setServiceIds[idx] = setServiceIds[setServiceIds.length - 1];
+        setServiceIds.pop();
+
+        emit ServiceUnstaked(serviceId, msg.sender, amount, tsStart);
```
mariapiamo, line 348: can we emit also service multisig address and nonce?

contracts/staking/ServiceStakingToken.sol
```diff
@@ -0,0 +1,125 @@
...
+
+    // ServiceRegistryTokenUtility address
+    address public immutable serviceRegistryTokenUtility;
+    // Security token address for staking corresponding to the service deposit token
+    address public immutable stakingToken;
+
+    /// @dev ServiceStakingToken constructor.
+    /// @param _apy Staking APY (in single digits).
+    /// @param _minStakingDeposit Minimum staking deposit for a service to be eligible to stake.
+    /// @param _stakingRatio Staking ratio: number of seconds per nonce (in 18 digits).
```
mariapiamo, line 48: livenessRatio ?  and it should be number of nonces for seconds

contracts/staking/ServiceStakingToken.sol
```diff
@@ -0,0 +1,125 @@
...
+    // Minimum number of token decimals
+    uint8 public constant MIN_DECIMALS = 4;
+
+    // ServiceRegistryTokenUtility address
+    address public immutable serviceRegistryTokenUtility;
+    // Security token address for staking corresponding to the service deposit token
+    address public immutable stakingToken;
+
+    /// @dev ServiceStakingToken constructor.
+    /// @param _apy Staking APY (in single digits).
```
mariapiamo, line 46: rewardsPerSecond? for consistency

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,384 @@
...
+    uint256 public minBalance;
+    // Mapping of serviceId => staking service info
+    mapping (uint256 => ServiceInfo) public mapServiceInfo;
+    // Set of currently staking serviceIds
+    uint256[] public setServiceIds;
+
+    /// @dev ServiceStakingBase constructor.
+    /// @param _rewardsPerSecond Staking rewards per second (in single digits).
+    /// @param _minStakingDeposit Minimum staking deposit for a service to be eligible to stake.
+    /// @param _livenessRatio Staking ratio: number of seconds per nonce (in 18 digits).
```
mariapiamo, line 100:
```suggestion
    /// @param _livenessRatio Liveness ratio: number of nonces per second (in 18 digits).
```

contracts/staking/ServiceStaking.sol
```diff
@@ -0,0 +1,47 @@
...
+
+/// @title ServiceStakingToken - Smart contract for staking a service by its owner when the service has an ETH as the deposit
+/// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
+/// @author Andrey Lebedev - <andrey.lebedev@valory.xyz>
+/// @author Mariapia Moscatiello - <mariapia.moscatiello@valory.xyz>
+contract ServiceStaking is ServiceStakingBase {
+    /// @dev ServiceStaking constructor.
+    /// @param _apy Staking APY (in single digits).
+    /// @param _minStakingDeposit Minimum staking deposit for a service to be eligible to stake.
+    /// @param _stakingRatio Staking ratio: number of seconds per nonce (in 18 digits).
```
mariapiamo, line 14: as for the other livenessRatio and rewardsPerSecond would be more consistent

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,384 @@
...
+                // Calculate the staking nonce ratio
+                // Get the last service checkpoint: staking start time or the global checkpoint timestamp
+                uint256 serviceCheckpoint = tsCheckpointLast;
+                // Adjust the service checkpoint time if the service was staking less than the current staking period
+                if (curInfo.tsStart > tsCheckpointLast) {
+                    serviceCheckpoint = curInfo.tsStart;
+                }
+                // Calculate the liveness ratio in 1e18 value
+                uint256 ratio;
+                // If the checkpoint was called in the exactly same block, the ratio is zero
```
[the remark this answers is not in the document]
kupermind: I don't see the issue here, as the last nonce will be registered at the moment of the checkpoint, so if more nonces are going to be in that block, they will be picked up the next time.

contracts/staking/ServiceStakingBase.sol
kupermind: yes

contracts/staking/ServiceStakingBase.sol
kupermind: yes

contracts/staking/ServiceStakingBase.sol
kupermind: why?

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    /// @dev Withdraws the reward amount to a service owner.
+    /// @param to Address to.
+    /// @param amount Amount to withdraw.
+    function _withdraw(address to, uint256 amount) internal virtual {}
+
+    /// @dev Stakes the service.
+    /// @param serviceId Service Id.
+    function stake(uint256 serviceId) external {
```
77ph, line 133: revert() if availableRewards is zero?

kupermind: I disagree, services can stake before rewards appear such that they are already in the staking accounting

kupermind: Nobody promised any rewards in the first place.

77ph: IMO: It would seem more logical to me that staking would be allowed when availableRewards > 0. Rewards are paid from these according to the scheme in the current contract. While availableRewards >0, you can still add stackers. When all funds are exhausted during the next rewards distribution cycle, the contract goes on "pause" (nobody can stake). And so on until the contract is replenished via donate ETH/Token. Then the cycle resumes.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -0,0 +1,400 @@
...
+            revert NoRewardsAvailable();
+        }
+
+        // Check for the maximum number of staking services
+        uint256 numStakingServices = setServiceIds.length;
+        if (numStakingServices == maxNumServices) {
+            revert MaxNumServicesReached(maxNumServices);
+        }
+
+        // Check the service conditions for staking
```
DavidMinarsch, line 160: I notice a nice optional feature which I'd like to support from the beginning: a staking contract can optionally specify an exact agentID and hash for the service that it requires to be staked. Then here, we check the agentId and service hash as well if and only if the hash is provided in the staking contract.

This would allow staking contracts to express exactly what service they want at which version.

This also raises the question if we should allow the staking contract to specify the number of agent instances and threshold. 

These four things together tie down the exact configuration.

contracts/staking/ServiceStakingBase.sol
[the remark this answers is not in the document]
DavidMinarsch: I guess @mariapiamo's point requires >=  rather than >?

contracts/staking/ServiceStakingBase.sol
DavidMinarsch: Is reverse loop more expensive? Makes sense to have the first staked service have this privilige.

contracts/staking/ServiceStaking.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.21;
+
+import {ServiceStakingBase} from "./ServiceStakingBase.sol";
+
+/// @title ServiceStakingToken - Smart contract for staking a service by its owner when the service has an ETH as the deposit
+/// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
+/// @author Andrey Lebedev - <andrey.lebedev@valory.xyz>
+/// @author Mariapia Moscatiello - <mariapia.moscatiello@valory.xyz>
+contract ServiceStaking is ServiceStakingBase {
```
DavidMinarsch, line 10: Maybe call it ServiceStakingNativeToken?

contracts/staking/ServiceStakingToken.sol
```diff
@@ -0,0 +1,116 @@
...
+    )
+        ServiceStakingBase(_maxNumServices, _rewardsPerSecond, _minStakingDeposit, _livenessRatio, _serviceRegistry)
+    {
+        // Initial checks
+        if (_stakingToken == address(0) || _serviceRegistryTokenUtility == address(0)) {
+            revert ZeroAddress();
+        }
+
+        stakingToken = _stakingToken;
+        serviceRegistryTokenUtility = _serviceRegistryTokenUtility;
```
DavidMinarsch, line 67: Should also check that the staking token matches the one used for security deposit? I mean in principle one could break this symmetry. But I think we should then have a different contract. Symmetry is more "expected"

contracts/staking/ServiceStakingToken.sol
```diff
@@ -0,0 +1,116 @@
...
+    /// @param serviceId Service Id.
+    function _checkTokenStakingDeposit(uint256 serviceId, uint256) internal view override {
+        // Get the service staking token and deposit
+        (address token, uint96 stakingDeposit) =
+            IServiceTokenUtility(serviceRegistryTokenUtility).mapServiceIdTokenDeposit(serviceId);
+
+        // The staking token must match the contract token
+        if (stakingToken != token) {
+            revert WrongStakingToken(stakingToken, token);
+        }
```
DavidMinarsch, line 80: Also in constructor

DavidMinarsch (commented):
Nice progress!

My biggest concern is now what I mentioned this morning: with this implementation it can happen that frequent staking/unstaking/checkpoint calls leads to "unjust" reward allocations simply because the liveness check is missed due to non-equally spaced transactions.

There's different ways to solve this:
1) we can say checkpoints can only happen every x time period. If people unstake before they get by default nothing for the rolling period.
2) we can somehow smoothen the liveness calculation across checkpoints. @mariapiamo might have ideas there?
3) other ideas...

contracts/staking/ServiceStakingBase.sol
kupermind: ok that's nice

contracts/staking/ServiceStakingToken.sol
kupermind: We check that for each service itself. ServiceRegistryTokenUtility is not bound to a specific custom token.

kupermind:
@DavidMinarsch we came to the conclusion that we would like to define a verification time window, say 1 day, and if the service staked longer than that, even if the checkpoint happened every block and the service had a lot of activity in a couple of blocks, the service activity would still be accounted for during the whole day period.

Restricting checkpoints is not very convenient, as it would potentially create a crowd of services trying to unstake, or complain that they have not grabbed their last share of a stake.

77ph (approved):

mariapiamo (approved):

PR #120, opened 2023-10-03
Service staking refactor

kupermind:
- Service staking implementation accounting for latest suggestions;
- Adding more tests.

### DIFF
--- a/contracts/staking/ServiceStakingBase.sol
+++ b/contracts/staking/ServiceStakingBase.sol
@@ -101,12 +101,12 @@ abstract contract ServiceStakingBase is IErrorsRegistries {
         uint256 livenessPeriod;
         // Liveness ratio in the format of 1e18
         uint256 livenessRatio;
+        // Number of agent instances in the service
+        uint256 numAgentInstances;
         // Optional agent Ids requirement
         uint256[] agentIds;
         // Optional service multisig threshold requirement
         uint256 threshold;
-        // Optional service number of agent instances requirement
-        uint256 numAgentInstances;
         // Optional service configuration hash requirement
         bytes32 configHash;
     }
@@ -130,10 +130,10 @@ abstract contract ServiceStakingBase is IErrorsRegistries {
     uint256 public immutable livenessPeriod;
     // Liveness ratio in the format of 1e18
     uint256 public immutable livenessRatio;
+    // Number of agent instances in the service
+    uint256 public immutable numAgentInstances;
     // Optional service multisig threshold requirement
     uint256 public immutable threshold;
-    // Optional service number of agent instances requirement
-    uint256 public immutable numAgentInstances;
     // Optional service configuration hash requirement
     bytes32 public immutable configHash;
     // ServiceRegistry contract address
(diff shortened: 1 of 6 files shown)


contracts/staking/ServiceStakingBase.sol
```diff
@@ -191,7 +281,6 @@ abstract contract ServiceStakingBase is IErrorsRegistries {
     /// @param eligibleServiceIds Service Ids eligible for rewards.
     /// @param eligibleServiceRewards Corresponding rewards for eligible service Ids.
     /// @param serviceIds All the staking service Ids.
-    /// @param serviceNonces Current service nonces.
```
mariapiamo, line 194: why this comment was removed?

contracts/staking/ServiceStakingBase.sol
```diff
@@ -202,96 +291,115 @@ abstract contract ServiceStakingBase is IErrorsRegistries {
...
+
+                    // Get current service multisig nonce
+                    serviceNonces[i] = IMultisig(curInfo.multisig).nonce();
+
+                    // Calculate the liveness nonce ratio
+                    // Get the last service checkpoint: staking start time or the global checkpoint timestamp
+                    uint256 serviceCheckpoint = tsCheckpointLast;
+                    uint256 tsStart = curInfo.tsStart;
+                    // Adjust the service checkpoint time if the service was staking less than the current staking period
+                    if (tsStart > serviceCheckpoint) {
```
mariapiamo, line 330: With this approach, we are not forcing a service to be staked for the entire liveness period. As we did previously, service multisigs that pass the liveness check will be considered for incentives, even if they have been staked for less than the liveness period. Is that correct?

contracts/staking/ServiceStakingBase.sol
```diff
@@ -202,96 +291,115 @@ abstract contract ServiceStakingBase is IErrorsRegistries {
...
+                        serviceCheckpoint = tsStart;
+                    }
+                    // Calculate the liveness ratio in 1e18 value
+                    uint256 ratio;
+                    // If the checkpoint was called in the exactly same block, the ratio is zero
```
mariapiamo, line 335: If the checkpoint was called in the exactly same block, we do not reach this part. No?

kupermind: I think it is possible if someone calls staking and checkpoint in the same block

contracts/staking/ServiceStakingBase.sol
kupermind: It was from the previous PR when this was considered obsolete, I've changed that now.

contracts/staking/ServiceStakingBase.sol
kupermind: That is correct, as the checkpoint is a centralization point and everybody eligible must account for their rewards.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -97,37 +146,56 @@ abstract contract ServiceStakingBase is IErrorsRegistries {
...
+        if (size > 0) {
+            for (uint256 i = 0; i < size; ++i) {
+                // Agent Ids must be unique and in ascending order
+                if (_stakingParams.agentIds[i] <= agentId) {
+                    revert WrongAgentId(_stakingParams.agentIds[i]);
+                }
+                agentId = _stakingParams.agentIds[i];
+                agentIds.push(agentId);
+            }
+        }
```
DavidMinarsch, line 195: Isn't this guaranteed by the registry already?

kupermind: Staking service agent Ids must match the staking contract configuration. We want to ensure at least the minimum order requirement.

77ph (approved):

PR #125, opened 2023-10-09
refactor and test: adding foundry tests

kupermind:
- Modifying code a little bit to account for the L2 contracts that don't have the component / agent information from the service.
- Adding foundry tests;
- Adding tests of modifying any nonce of any multisig (`SafeNonceLib` contract);
- Adding to the audit remarks.


contracts/staking/ServiceStakingBase.sol
```diff
@@ -13,11 +13,6 @@ interface IMultisig {
 
 // Service Registry interface
 interface IService {
-    enum UnitType {
-        Component,
-        Agent
-    }
-
```
kupermind, line 20: We can't use the `getUnitIdsOfService()` function as it's available only for L1. Instead, we need to get the full service data and extract `agentIds` from it.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -254,67 +248,65 @@ abstract contract ServiceStakingBase is ERC721TokenReceiver, IErrorsRegistries {
         }
 
         // Check the service conditions for staking
-        (uint96 stakingDeposit, address multisig, bytes32 hash, uint256 agentThreshold, uint256 maxNumInstances, , uint8 state) =
-            IService(serviceRegistry).mapServices(serviceId);
+        IService.Service memory service = IService(serviceRegistry).getService(serviceId);
```
kupermind, line 251: Get the full service and extract the necessary data from it.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -565,4 +556,11 @@ abstract contract ServiceStakingBase is ERC721TokenReceiver, IErrorsRegistries {
             }
         }
     }
+
+    /// @dev Checks if the service is staked.
+    /// @param serviceId.
+    /// @return True, if the service is staked.
+    function isServiceStaked(uint256 serviceId) external view returns (bool) {
+        return mapServiceInfo[serviceId].tsStart > 0;
+    }
```
kupermind, line 565: Adding a simple function to check if the service is staked.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -162,19 +157,22 @@ abstract contract ServiceStakingBase is ERC721TokenReceiver, IErrorsRegistries {
...
         if (_stakingParams.maxNumServices == 0 || _stakingParams.rewardsPerSecond == 0 ||
-            _stakingParams.minStakingDeposit == 0 || _stakingParams.livenessPeriod == 0 ||
-            _stakingParams.livenessRatio == 0 || _stakingParams.numAgentInstances == 0) {
+            _stakingParams.livenessPeriod == 0 || _stakingParams.livenessRatio == 0 ||
+            _stakingParams.numAgentInstances == 0) {
             revert ZeroValue();
         }
+        if (_stakingParams.minStakingDeposit < 2) {
+            revert LowerThan(_stakingParams.minStakingDeposit, 2);
+        }
```
kupermind, line 175: This is done such that somehow the minimum staking deposit is not set to 1 (although very unlikely). If that happens, the NativeToken and Token contracts will be undistinguishable, and ERC20-based service will be possible to stake at the NativeToken staking contract.

audits/internal4/README.md
```diff
@@ -138,7 +138,24 @@ We can call execTransaction from another contract as owner in loop.
...
+
+### Found during fuzzing. Updated 10-10-23
+#### Library that changes the value of nonce for legitimate multisig
+It turned out that if you know the architecture of Saf1 1.3.x it is not difficult to write a library for changing nonce. <br>
+Two conclusions follow from this: <br>
+1. The initial assumption that `nonce` can only increase is wrong. So, serviceNonces[i] - curInfo.nonce must be checked by > 0. <br>
+Else without it: ratio = ((serviceNonces[i] - curInfo.nonce) * 1e18) / ts; => DOS (revert) in _calculateStakingRewards() with any serviceNonces[i] - curInfo.nonce < 0 <br>
+2. ref: General considerations #2. We cannot guarantee the absence of manipulation nonce through any checks within the contract. <br>
+A reasonable proposal in this case is to provide the opportunity to add a blacklist of services when deploying a staking contract. <br>
+This will make it possible to blacklist services with a bad reputation (which is not part of the protocol, but determined by the community). <br>
```
DavidMinarsch, line 152: That's not worth the hassle. Then you need governance for this (in some form). Furthermore, creating a new service that makes the same exploit is possible.

audits/internal4/README.md
```diff
@@ -138,7 +138,24 @@ We can call execTransaction from another contract as owner in loop.
...
+
+#### Simple function needed `isServiceStaked`
+
+#### _stakingParams.minStakingDeposit
+minStakingDeposit must be greater than 1 otherwise it causes confusion with the predefined value of 1.
```
DavidMinarsch, line 157: why is the pre-defined value not 0?

DavidMinarsch:
Here https://github.com/valory-xyz/autonolas-registries/blob/fuzzing/contracts/staking/ServiceStakingBase.sol#L363-L381 you write and read from storage, would use memory instead for the nonce of i

DavidMinarsch:
Please extract this (https://github.com/valory-xyz/autonolas-registries/blob/fuzzing/contracts/staking/ServiceStakingBase.sol#L374-L380) into a private, overridable method "_compute_liveness_ratio". Then any contract building on this can simply override this implementation to offer a different liveness check.

audits/internal4/README.md
77ph: For ServiceManagerToken.sol
 // Wrap bonds with the BOND_WRAPPER value for the original ServiceRegistry 
contractagentParams[i].bond = BOND_WRAPPER;
in this sense "pre-defined"
I think @kupermind  can explain better than me

audits/internal4/README.md
77ph: In essence, this is an attempt to solve, not by technical means, what cannot be solved within "on-chain" approach. 
In the end, we can simply ignore it to begin with if/until the problem is not widespread. 
What is needed here is some kind of offline reputation system. Made by "community".
As I see it.

audits/internal4/README.md
kupermind: This is done such that somehow the minimum staking deposit is not set to 1 (although very unlikely). If that happens, the NativeToken and Token contracts will be undistinguishable, and ERC20-based service will be possible to stake at the NativeToken staking contract.

kupermind:
Note: the extraction of the liveness calculation function is done in the following PR.

mariapiamo (approved):

PR #126, opened 2023-10-10
refactor and test: update same address multisig implementation

kupermind:
- Update `GnosisSafeSameAddressMultisig` implementation to check for the proxy;
- Add / update tests.

### DIFF
--- a/contracts/multisigs/GnosisSafeSameAddressMultisig.sol
+++ b/contracts/multisigs/GnosisSafeSameAddressMultisig.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.15;
+pragma solidity ^0.8.21;
 
 // Gnosis Safe Master Copy interface extracted from the mainnet: https://etherscan.io/address/0xd9db270c1b5e3bd161e8c8503c55ceabee709552#code#F6#L126
 interface IGnosisSafe {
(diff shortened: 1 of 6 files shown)


contracts/multisigs/GnosisSafeSameAddressMultisig.sol
```diff
@@ -42,6 +52,31 @@ contract GnosisSafeSameAddressMultisig {
...
+    /// @dev GnosisSafeSameAddressMultisig constructor.
+    /// @param _multisigProxyHashes Multisig proxy hashes.
+    constructor(bytes32[] memory _multisigProxyHashes) {
+        // There must be at least one multisig proxy hash
+        uint256 size = _multisigProxyHashes.length;
+        if (size == 0) {
+            revert ZeroValue();
+        }
+
+        // Record provided multisig proxy bytecode hashes
```
mariapiamo, line 67: we need to update the deployment scripts with these hashes

test/GnosisSafeSameAddressMultisig.js
```diff
@@ -39,6 +46,19 @@ describe("GnosisSafeSameAddressMultisig", function () {
...
+            await expect(
+                GnosisSafeSameAddressMultisig.deploy([])
+            ).to.be.revertedWithCustomError(gnosisSafeSameAddressMultisig, "ZeroValue");
+
+            const bytes32Zero = "0x" + "0".repeat(64);
+            await expect(
+                GnosisSafeSameAddressMultisig.deploy([bytes32Zero])
+            ).to.be.revertedWithCustomError(gnosisSafeSameAddressMultisig, "ZeroValue");
+        });
+
```
mariapiamo, line 61: we should also update the tests regarding the GnosisSafeSameAddressMultisig.create() method, no?

kupermind: There is one in the `serviceStaking` testing

mariapiamo (approved):

PR #127, opened 2023-10-11
setting one multisig proxy and updating deployment scripts

kupermind:
- Setting one multisig proxy bytecode hash;
- Updating deployment scripts.


mariapiamo (approved):

mariapiamo (commented):
We should also modify https://github.com/valory-xyz/autonolas-registries/blob/main/scripts/deployment/README.md#deployment-of-supplemental-contracts and https://github.com/valory-xyz/autonolas-registries/blob/main/docs/deploymentL2.md to link the deployment of GnosisSafeSameAddressMultisig contract

mariapiamo (approved):

PR #135, opened 2023-10-24
refactor and test: service staking optimization

kupermind:
- Optimizing service staking contract logic;
- Adding tests.


contracts/staking/ServiceStakingBase.sol
```diff
@@ -352,16 +352,6 @@ abstract contract ServiceStakingBase is ERC721TokenReceiver, IErrorsRegistries {
...
-        // Get the service Ids set length
-        uint256 size = setServiceIds.length;
-        serviceIds = new uint256[](size);
-
-        // Record service Ids
-        for (uint256 i = 0; i < size; ++i) {
-            // Get current service Id
-            serviceIds[i] = setServiceIds[i];
-        }
-
```
kupermind, line 364: Don't allocate `serviceIds` if the rewards calculation is not going to be successful.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -508,7 +500,12 @@ abstract contract ServiceStakingBase is ERC721TokenReceiver, IErrorsRegistries {
         }
 
         // Call the checkpoint
-        (uint256[] memory serviceIds, , , , , ) = checkpoint();
+        (uint256[] memory serviceIds, , , , , bool success) = checkpoint();
+
+        // If the checkpoint was not successful, the serviceIds set is not returned and needs to be allocated
+        if (!success) {
+            serviceIds = getServiceIds();
+        }
```
kupermind, line 508: Checkpoint now might not return `serviceIds` in memory if it was not successful. So in that case it must be obtained separately.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -562,26 +559,44 @@ abstract contract ServiceStakingBase is ERC721TokenReceiver, IErrorsRegistries {
...
+                    reward += eligibleServiceRewards[i];
                 }
+                break;
             }
         }
     }
 
+    /// @dev Gets staked service Ids.
+    /// @return serviceIds Staked service Ids.
+    function getServiceIds() public view returns (uint256[] memory serviceIds) {
```
kupermind, line 578: Adding the function that returns current staking `serviceIds`. The default ABI getter would not be enough as it does not return all of the Ids and the length, but just a single element with the provided index.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -562,26 +559,44 @@ abstract contract ServiceStakingBase is ERC721TokenReceiver, IErrorsRegistries {
...
+    /// @return isStaked True, if the service is staked.
+    function isServiceStaked(uint256 serviceId) external view returns (bool isStaked) {
+        isStaked = (mapServiceInfo[serviceId].tsStart > 0);
+    }
+
+    /// @dev Gets the next reward checkpoint timestamp.
+    /// @return tsNext Next reward checkpoint timestamp.
+    function getNextRewardCheckpointTimestamp() external view returns (uint256 tsNext) {
+        // Last checkpoint timestamp plus the liveness period
+        tsNext = tsCheckpoint + livenessPeriod;
```
kupermind, line 600: Getter of the next reward checkpoint timestamp - the agent can always compare with the `block.timestamp` and decide whether to call a `checkpoint()` or skip for now.

mariapiamo (approved):

PR #139, opened 2023-11-03
feat: Performing service downtime check

kupermind:
- Performing service inactivity check;
- When merged, service staking programs need to be updated as well.


77ph (approved):

mariapiamo:
I was starting some monitoring, the checkpoint event is very low informative: it does not provide the serviceIds with rewards allocation and the corresponding amount. It is necessary to update the checkpoint event with the previous info

contracts/staking/ServiceStakingBase.sol
```diff
@@ -235,6 +259,12 @@ abstract contract ServiceStakingBase is ERC721TokenReceiver, IErrorsRegistries {
             revert NoRewardsAvailable();
         }
 
+        // Check if the evicted service has not yet unstaked
+        ServiceInfo storage sInfo = mapServiceInfo[serviceId];
+        if (sInfo.tsStart > 0) {
```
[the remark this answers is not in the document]
kupermind: When the service is evicted, it is removed from the staking set, but its info is still in the map. So if the service tries to stake again, and its tsStart is not zero, the service is not unstaked yet.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -287,7 +317,6 @@ abstract contract ServiceStakingBase is ERC721TokenReceiver, IErrorsRegistries {
         _checkTokenStakingDeposit(serviceId, service.securityDeposit);
 
         // ServiceInfo struct will be an empty one since otherwise the safeTransferFrom above would fail
-        ServiceInfo storage sInfo = mapServiceInfo[serviceId];
```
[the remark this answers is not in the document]
kupermind: It is not removed, just moved up the code to get the struct value earlier.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -541,38 +593,78 @@ abstract contract ServiceStakingBase is ERC721TokenReceiver, IErrorsRegistries {
...
+        ServiceInfo memory sInfo = mapServiceInfo[serviceId];
+        reward = sInfo.reward;
+
+        // Check if the service is staked
+        if (sInfo.tsStart == 0) {
+            revert ServiceNotStaked(serviceId);
+        }
+
+        reward += calculateServiceStakingLastReward(serviceId);
+    }
```
kupermind, line 666: @mariapiamo it could be very intense to try to emit all the services and rewards during the Checkpoint() emit. We have added one more view reward function - one can always check a reward during a specific period specifying the block number. So there is a function to calculate the last reward, and there's a function to calculate the overall reward.

contracts/staking/ServiceStakingBase.sol
```diff
@@ -393,11 +424,12 @@ abstract contract ServiceStakingBase is ERC721TokenReceiver, IErrorsRegistries {
                 // Record the reward for the service if it has provided enough transactions
                 if (ratioPass) {
                     // Calculate the reward up until now and record its value for the corresponding service
-                    uint256 reward = rewardsPerSecond * ts;
-                    totalRewards += reward;
-                    eligibleServiceRewards[numServices] = reward;
+                    eligibleServiceRewards[numServices] = rewardsPerSecond * ts;
+                    totalRewards += eligibleServiceRewards[numServices];
```
DavidMinarsch, line 428: Wasn't the old implementation more gas efficient?

contracts/staking/ServiceStakingBase.sol
```diff
@@ -499,6 +540,13 @@ abstract contract ServiceStakingBase is ERC721TokenReceiver, IErrorsRegistries {
             revert OwnerOnly(msg.sender, sInfo.owner);
         }
 
+        // Check that the service has staked long enough, or if there are no rewards left
+        uint256 tsStart = sInfo.tsStart;
+        uint256 ts = block.timestamp - tsStart;
+        if (ts <= maxAllowedInactivity && availableRewards > 0) {
+            revert NotEnoughTimeStaked(serviceId, ts, maxAllowedInactivity);
```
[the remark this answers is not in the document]
DavidMinarsch: For a multi-agent service yes. For a single agent service no. Let's discuss ..

contracts/staking/ServiceStakingBase.sol
```diff
@@ -519,16 +567,20 @@ abstract contract ServiceStakingBase is ERC721TokenReceiver, IErrorsRegistries {
         // Get the unstaked service data
         uint256 reward = sInfo.reward;
         uint256[] memory nonces = sInfo.nonces;
-        uint256 tsStart = sInfo.tsStart;
         address multisig = sInfo.multisig;
 
+        // TODO: Add inactivity to the global map of inactivities, in order to never let stake if that limit has been breached
```
DavidMinarsch, line 572: We didn't discuss this. Right now evicted services will be slashed so there's no reason to perma-ban them.

DavidMinarsch: (which is not sybil resistant anyway)

contracts/staking/ServiceStakingBase.sol
```diff
@@ -541,38 +593,78 @@ abstract contract ServiceStakingBase is ERC721TokenReceiver, IErrorsRegistries {
...
         for (uint256 i = 0; i < numServices; ++i) {
             // Get the service index in the eligible service set and calculate its latest reward
             if (eligibleServiceIds[i] == serviceId) {
                 // If total allocated rewards are not enough, adjust the reward value
                 if (totalRewards > lastAvailableRewards) {
-                    reward += (eligibleServiceRewards[i] * lastAvailableRewards) / totalRewards;
+                    reward = (eligibleServiceRewards[i] * lastAvailableRewards) / totalRewards;
                 } else {
-                    reward += eligibleServiceRewards[i];
+                    reward = eligibleServiceRewards[i];
```
DavidMinarsch, line 645: why the change?

contracts/staking/ServiceStakingBase.sol
```diff
@@ -541,38 +593,78 @@ abstract contract ServiceStakingBase is ERC721TokenReceiver, IErrorsRegistries {
...
 
-    /// @dev Calculates service staking reward at current timestamp.
-    /// @param serviceId Service Id.
-    /// @return reward Service reward.
-    function calculateServiceStakingReward(uint256 serviceId) external view returns (uint256 reward) {
-        // Get current service reward
-        ServiceInfo memory sInfo = mapServiceInfo[serviceId];
-        reward = sInfo.reward;
+    /// @dev Evicts the service due to its extended inactivity.
+    function evict(uint256 serviceId) external {
```
DavidMinarsch, line 597: As part of checkpoint we should evict all that match the eviction condition imo.

contracts/staking/ServiceStakingBase.sol
kupermind: Yes, we are just addressing comments for calculating reward for the current period, as Mariapia stated we wanted to have more information about rewards. I have a comment just above. Also, it's a view function that is not used on-chain by the staking contract.

contracts/staking/ServiceStakingBase.sol
kupermind: It was left from the previous implementation.

contracts/staking/ServiceStakingBase.sol
kupermind: In order to separate reward calculation for the current checkpoint and the accumulated one - see above.

contracts/staking/ServiceStakingBase.sol
kupermind: This is a bigger change that can be applied in a next PR IMO.

contracts/staking/ServiceStakingBase.sol
mariapiamo: I see that can be very intense and I know about that view method. But rn there is no way to use that method to create a query

kupermind: Does the separation of view functions into `calculateServiceStakingReward()` and `calculateServiceStakingLastReward()` work for that purpose?

mariapiamo: No, we need events. It is super intense trying to emit all the services and rewards during the Checkpoint() ?

contracts/staking/ServiceStakingBase.sol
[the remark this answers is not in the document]
DavidMinarsch: Let's add a bit more of this comment in code to make it clearer

contracts/staking/ServiceStakingBase.sol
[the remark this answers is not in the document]
DavidMinarsch: I think we should leave it as is and just add a note on the top of the contract to notice of this limitation.

contracts/staking/ServiceStakingBase.sol
DavidMinarsch: @mariapiamo why do you want so much info in the events? We can call a view function just fine?

mariapiamo: @DavidMinarsch rn there is no way to use that method to create sql queries

contracts/staking/ServiceStakingBase.sol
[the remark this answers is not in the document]
kupermind: Modified in #141

contracts/staking/ServiceStakingBase.sol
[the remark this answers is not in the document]
kupermind: Added in #141

PR #186, opened 2024-07-16
refactor: account for staking contract deposit value

kupermind:
- Account for staking contract deposit value.

### DIFF
--- a/contracts/staking/StakingVerifier.sol
+++ b/contracts/staking/StakingVerifier.sol
@@ -22,6 +22,9 @@ interface IStaking {
     /// @dev Gets service registry token utility address.
     /// @return Service registry token utility address.
     function serviceRegistryTokenUtility() external view returns(address);
+
+    /// @dev Minimum service staking deposit value required for staking.
+    function minStakingDeposit() external view returns(uint256);
 }
 
 /// @dev Provided zero address.
@@ -52,8 +55,8 @@ contract StakingVerifier {
     event OwnerUpdated(address indexed owner);
     event SetImplementationsCheck(bool setCheck);
     event ImplementationsWhitelistUpdated(address[] implementations, bool[] statuses, bool setCheck);
-    event StakingLimitsUpdated(uint256 rewardsPerSecondLimit, uint256 timeForEmissionsLimit, uint256 _numServicesLimit,
-        uint256 emissionsLimit);
+    event StakingLimitsUpdated(uint256 rewardsPerSecondLimit, uint256 timeForEmissionsLimit, uint256 numServicesLimit,
+        uint256 emissionsLimit, uint256 depositFactor);
 
     // OLAS token address
     address public immutable olas;
@@ -70,6 +73,8 @@ contract StakingVerifier {
     uint256 public numServicesLimit;
     // Emissions per service limit
     uint256 public emissionsLimit;
+    // Deposit factor
+    uint256 public depositFactor;
     // Contract owner address
     address public owner;
     // Flag to check for the implementation address whitelisting status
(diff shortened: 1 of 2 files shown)


contracts/staking/StakingVerifier.sol
```diff
@@ -279,36 +312,38 @@ contract StakingVerifier {
...
         // Check the contract ownership
         if (owner != msg.sender) {
             revert OwnerOnly(owner, msg.sender);
         }
 
         // Zero values check
-        if (_rewardsPerSecondLimit == 0 || _timeForEmissionsLimit == 0 || _numServicesLimit == 0) {
+        if (_minStakingDepositLimit == 0 || _timeForEmissionsLimit == 0 || _numServicesLimit == 0 || _apyLimit == 0) {
             revert ZeroValue();
         }
```
DavidMinarsch, line 334: where's the setter for _minStakingDepositLimit?

kupermind: Constructor and `changeStakingLimits()` function, it just replaces `rewardsPerSecondLimit` everywhere.

DavidMinarsch: Yes but see below,l we are not able to update _minStakingDepositLimit as the setter is missing

yarn.lock
```diff
@@ -686,53 +686,53 @@
     "@nodelib/fs.scandir" "2.1.5"
     fastq "^1.6.0"
 
-"@nomicfoundation/edr-darwin-arm64@0.3.8":
```
DavidMinarsch, line 689: Do any of the yarn updates cause new dependencies / dependencies being updated at the contract level. That should not be the case as they've already been audited

contracts/staking/StakingVerifier.sol
```diff
@@ -225,10 +241,27 @@ contract StakingVerifier {
...
+        uint256 rewardsPerYear = IStaking(instance).rewardsPerSecond() * ONE_YEAR;
+        // Calculate current APY in 1e18 format
+        uint256 apy = (rewardsPerYear * 1e18) / minStakingDeposit;
+
+        // Compare APY with the limit
+        if (apy > apyLimit) {
+            return false;
+        }
+
+        // Check for time for emissions
```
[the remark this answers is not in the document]
kupermind: No, it's related to `emissionsAmount` of instance that requests that amount from the incentive dispenser.

yarn.lock
kupermind: No, this is just part of the hardhat upgrade. Nothing code related.

77ph (approved):

contracts/staking/StakingFactory.sol
```diff
@@ -296,29 +299,22 @@ contract StakingFactory {
...
-    function verifyInstance(address instance) public view returns (bool) {
-        // Get proxy instance params
+    function verifyInstance(address instance) external view returns (bool) {
         InstanceParams storage instanceParams = mapInstanceParams[instance];
-        address implementation = instanceParams.implementation;
-
-        // Check that the implementation corresponds to the proxy instance
-        if (implementation == address(0)) {
-            return false;
-        }
```
DavidMinarsch, line 311: Why was this removed?

contracts/staking/StakingVerifier.sol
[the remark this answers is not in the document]
DavidMinarsch: Can we add the relevant getters so a future Verifier can also impose constrains on livenessPeriod?

contracts/staking/StakingFactory.sol
kupermind: This check is now useless, it was left there before the struct with implementation / deployer / isEnabled existed. The `isEnabled` is what matters, and if the `implementation` is zero, the `isEnabled` is false by default.

contracts/staking/StakingVerifier.sol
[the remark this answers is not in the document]
kupermind: You mean add the `livenessPeriodLimit` variable?

DavidMinarsch (approved):

77ph (approved):

PR #205, opened 2025-04-28
feat: multisig recovery module

kupermind:
- Multisig recovery module


77ph (approved):

contracts/multisigs/SafeMultisigWithRecoveryModule.sol
```diff
@@ -0,0 +1,107 @@
...
+
+            // Decode fallback handler and nonce
+            (fallbackHandler, nonce) = abi.decode(data, (address, uint256));
+        }
+
+        // Convert enableModule selector into bytes
+        bytes memory payload = bytes.concat(ENABLE_MODULE_SELECTOR);
+
+        // Encode the gnosis setup function parameters
+        bytes memory safeParams = abi.encodeWithSelector(SAFE_SETUP_SELECTOR, owners, threshold, recoveryModule,
```
DavidMinarsch, line 99: so `recoveryModule` replaces `to` here https://github.com/valory-xyz/autonolas-registries/blob/main/contracts/multisigs/GnosisSafeMultisig.sol#L103C13-L103C15

contracts/multisigs/SafeMultisigWithRecoveryModule.sol
```diff
@@ -0,0 +1,107 @@
...
+        safeProxyFactory = _safeProxyFactory;
+        recoveryModule = _recoveryModule;
+    }
+
+    /// @dev Creates a Safe multisig.
+    /// @param owners Set of multisig owners.
+    /// @param threshold Number of required confirmations for a multisig transaction.
+    /// @param data Encoded data related to the creation of a chosen multisig.
+    /// @return multisig Address of a created multisig.
+    function create(
```
DavidMinarsch, line 70: any reason we don't do a symmetric "parseData" implementation to this https://github.com/valory-xyz/autonolas-registries/blob/main/contracts/multisigs/GnosisSafeMultisig.sol#L45 ? would make comparisons trivial

contracts/multisigs/RecoveryModule.sol
```diff
@@ -0,0 +1,403 @@
...
+        // Check service owner
+        if (msg.sender != serviceOwner) {
+            revert OwnerOnly(msg.sender, serviceOwner);
+        }
+
+        // Check service state
+        (, address multisig, , , , , IServiceRegistry.ServiceState state) = IServiceRegistry(serviceRegistry).mapServices(serviceId);
+        if (state != IServiceRegistry.ServiceState.PreRegistration) {
+            revert WrongServiceState(uint8(state), serviceId);
+        }
```
DavidMinarsch, line 235: I am not sure why we restrict it like this. A typical lost state can involve where the service is "Deployed"

DavidMinarsch: In PreRegistration the service is owned by the service owner anyway. So why even use recovery?

contracts/multisigs/RecoveryModule.sol
```diff
@@ -0,0 +1,403 @@
...
+
+        emit AccessRecovered(msg.sender, serviceId);
+
+        _locked = 1;
+    }
+
+    /// @dev Updates and/or verifies the existent gnosis safe multisig for changed owners and threshold.
+    /// @notice This function operates with existent multisig proxy that is requested to be updated in terms of
+    ///         the set of owners' addresses and the threshold. There are two scenarios possible:
+    ///         1. The multisig proxy is already updated before reaching this function. Then the multisig address
```
DavidMinarsch, line 282: when is this the valid scenario?

contracts/multisigs/SafeMultisigWithRecoveryModule.sol
kupermind: To be honest there's no need for that. We can only parse `fallbackHandler` address and `nonce`. We don't want to parse any additional payload since we only allow the module related one. All the rest of params - those are useless as gas tokens do not exist for a long time already, and nobody is going to send any tokens during the Safe setup.

contracts/multisigs/RecoveryModule.sol
kupermind: We are recovering access to the Agent Safe (service multisig), that is controlled by agent instances, if agent instances keys are lost. We don't want the service owner to reclaim the multisig ownership when the service is still deployed. Also, unbonding is triggered by operators and not by agent instances, so even if agent instance keys are lost at that time, the unbonding takes place (operator keys are out of scope, as in majority of cases those are service owners). Thus, the service owner is able to change the configuration only when the service is fully terminated and unbonded.

At most we can add a valid recoverAccess state when the service is terminated but not fully unbonded if we project there could be cases that operators lose their keys and are unable to unbond.

contracts/multisigs/RecoveryModule.sol
kupermind: Please see this scenario (the service owner has updated the multisig before re-deployment) execution here:
https://github.com/valory-xyz/autonolas-registries/blob/586c632d2246f36d7eaabab46988a5d6cf676bf0/test/ServiceRegistry.js#L3838-L3863

kupermind: oh nice! code widgets now :)

kupermind: Just to add to that - when the module is enabled, service re-deployment becomes trivial as we are able to figure out all the new service config details and encode all the relevant data in-place, since the module does not need signatures for off-chain prepared transactions in order to execute them on behalf of the multisig.

contracts/multisigs/SafeMultisigWithRecoveryModule.sol
kupermind: FYI re: gas tokens, thanks @77ph for the link:
https://blog.openzeppelin.com/fundamentals-of-gas-tokens

I believe that part of the code is left even in latest version of Safe implementation for legacy purposes only.

DavidMinarsch: ok

mariapiamo (approved):

77ph (approved):

contracts/multisigs/RecoveryModule.sol
DavidMinarsch: "We are recovering access to the Agent Safe (service multisig), that is controlled by agent instances, if agent instances keys are lost." - that tends to be in deployed state!

"We don't want the service owner to reclaim the multisig ownership when the service is still deployed." Well in this case we want to. It's a tradeoff which by adopting the recovery module the service owner accepts.

"Also, unbonding is triggered by operators and not by agent instances, so even if agent instance keys are lost at that time, the unbonding takes place (operator keys are out of scope, as in majority of cases those are service owners)." - true. the question is what the threat model assumes here. We need to check with Jose if the service owner is the operator here. If that's the case then we are not constrained.

kupermind: We had a longer discussion today, and there was a concern that service owner does something with the multisig while the service is still deployed. The service owner can always terminate a service, so that's not a concern. We need to understand if operator, which might not be equal to the service owner, is a concern.

If operators do not unbond, the service multisig is never recoverable. We can then let the recovery take place when the service is terminated and not unbonded. If operators still don't unbond, the only way to recover agent bonds is via slashing.

For now we agreed that the operators not willing to unbond is out of scope, and the recovery happens on a pre-registration state. However, if we expect misbehaving operators, then we might reconsider recovery in terminated state as well. I'll let Jose comment from here.

PR #207, opened 2025-05-20
feat: first implementation of universal service hash updater

kupermind:
- First implementation of universal service hash updater.


mariapiamo (commented):
can we also add a getter providing the info of the operator able to perform the update hash for a specified service id?

contracts/utils/HashUpdater.sol
```diff
@@ -0,0 +1,103 @@
...
+
+/// @dev Account is unauthorized.
+/// @param account Account address.
+error UnauthorizedAccount(address account);
+
+/// @title HashUpdater - Metadata Hash Updater contract to manage actual / additional hashes
+/// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
+/// @author Andrey Lebedev - <andrey.lebedev@valory.xyz>
+/// @author Mariapia Moscatiello - <mariapia.moscatiello@valory.xyz>
+contract HashUpdater {
```
DavidMinarsch, line 47: How about `ComplementaryServiceMetadata`

That's descriptive of what it is

contracts/utils/HashUpdater.sol
```diff
@@ -0,0 +1,103 @@
...
+        address serviceOwner = IServiceRegistry(serviceRegistry).ownerOf(serviceId);
+        // Check for service owner
+        if (msg.sender != serviceOwner) {
+            // Get service multisig
+            (, address multisig, , , , , ) = IServiceRegistry(serviceRegistry).mapServices(serviceId);
+
+            // Check for service multisig address
+            if (msg.sender != multisig) {
+                revert UnauthorizedAccount(msg.sender);
+            }
```
DavidMinarsch, line 93: Imo this should be state dependent so that the service owner can update when they are in control and otherwise the agent service multisig itself can update.

kupermind: I believe this is exactly the case. When the service is not staked  - the service owner is always in control, otherwise it's multisig, as the staking contract being a service owner can't call this contract.

contracts/utils/HashUpdater.sol
```diff
@@ -0,0 +1,103 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.30;
+
+/// @dev Service Registry interface
+interface IServiceRegistry {
+    enum ServiceState {
```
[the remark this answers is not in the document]
kupermind: In the `mapServices()` function for now, but maybe we change it to `uint8` instead

77ph (approved):
forge scripts lgtm

contracts/utils/ComplementaryServiceMetadata.sol
```diff
@@ -0,0 +1,122 @@
...
+
+/// @dev Account is unauthorized.
+/// @param account Account address.
+error UnauthorizedAccount(address account);
+
+/// @title ComplementaryServiceMetadata - Complementary Service Metadata contract to manage complementary service hashes
+/// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
+/// @author Andrey Lebedev - <andrey.lebedev@valory.xyz>
+/// @author Mariapia Moscatiello - <mariapia.moscatiello@valory.xyz>
+contract ComplementaryServiceMetadata {
```
DavidMinarsch, line 47: Please let's make this a proxy so we can easily update implementation if need be. Governance via DAO

DavidMinarsch: Or maybe we don't need that because we can just create new singletons

contracts/utils/ComplementaryServiceMetadata.sol
```diff
@@ -0,0 +1,122 @@
...
+    string public constant VERSION = "0.1.0";
+
+    // Service Registry address
+    address public immutable serviceRegistry;
+
+    // Reentrancy lock
+    uint256 internal _locked = 1;
+
+    // Mapping of serviceId => actual metadata hash
+    mapping(uint256 => bytes32) public mapServiceHashes;
```
DavidMinarsch, line 60: imo we should apply the same sanitising for the ipfs string pattern we do in the service registry

contracts/utils/ComplementaryServiceMetadata.sol
kupermind: Correct, the contract is owner-less, let's keep it this way, since we can specify the singleton with complementary metadata in the service metadata config.

contracts/utils/ComplementaryServiceMetadata.sol
kupermind: We don't do anything specific there except for returning the URI with the prefix Do we want to use that here? And use `baseURI` from serviceRegistry?

DavidMinarsch: Yes let's return the full url composed from prefix etc

77ph (approved):

PR #208, opened 2025-05-22
refactor: access check is moved into its own view function

kupermind:
- Access check is moved into its own view function.


contracts/utils/ComplementaryServiceMetadata.sol
```diff
@@ -80,30 +80,43 @@ contract ComplementaryServiceMetadata {
         }
         _locked = 2;
 
+        // Check for msg.sender access to change hash
+        if (!isAbleChangeHash(msg.sender, serviceId)) {
```
kupermind, line 84: The initial logic is fully preserved.

77ph (approved):

PR #232, opened 2025-11-05
feat: Initial 8004 bridging implementation

kupermind:
- Initial 8004 bridging implementation, where `IdentityRegistryBridger.sol` acts as a bridge to ERC8004 IdentityRegistry, ensuring that  state between both Olas and ERC8004 registries stays synced.
- It will then further be linked with origin la OLAS AI agent registry `ServiceManager`, in order to have a seamless way of coupling agents to be ERC-8004 compatible.

### DIFF
--- a/scripts/deployment/l2/script_l2_01_whitelist_multisig_implementations.sh
+++ b/scripts/deployment/l2/script_l2_01_whitelist_multisig_implementations.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+
+# Check if $1 is provided
+if [ -z "$1" ]; then
+  echo "Usage: $0 <network>"
+  echo "Example: $0 eth_mainnet"
+  exit 1
+fi
+
+red=$(tput setaf 1)
+green=$(tput setaf 2)
+reset=$(tput sgr0)
+
+# Get globals file
+globals="$(dirname "$0")/globals_$1.json"
+if [ ! -f $globals ]; then
+  echo "${red}!!! $globals is not found${reset}"
+  exit 0
+fi
+
+# Read variables using jq
+useLedger=$(jq -r '.useLedger' $globals)
+derivationPath=$(jq -r '.derivationPath' $globals)
+chainId=$(jq -r '.chainId' $globals)
+networkURL=$(jq -r '.networkURL' $globals)
+
+serviceRegistryAddress=$(jq -r '.serviceRegistryAddress' $globals)
+safeMultisigWithRecoveryModuleAddress=$(jq -r '.safeMultisigWithRecoveryModuleAddress' $globals)
+recoveryModuleAddress=$(jq -r '.recoveryModuleAddress' $globals)
+
+# Check for Polygon keys only since on other networks those are not needed
+if [ $chainId == 137 ]; then
+  API_KEY=$ALCHEMY_API_KEY_MATIC
+  if [ "$API_KEY" == "" ]; then
+      echo "set ALCHEMY_API_KEY_MATIC env variable"
+      exit 0
+  fi
+elif [ $chainId == 80002 ]; then
+    API_KEY=$ALCHEMY_API_KEY_AMOY
+    if [ "$API_KEY" == "" ]; then
+        echo "set ALCHEMY_API_KEY_AMOY env variable"
+        exit 0
+    fi
+fi
+
+# Get deployer based on the ledger flag
+if [ "$useLedger" == "true" ]; then
+  walletArgs="-l --mnemonic-derivation-path $derivationPath"
+  deployer=$(cast wallet address $walletArgs)
+else
+  echo "Using PRIVATE_KEY: ${PRIVATE_KEY:0:6}..."
+  walletArgs="--private-key $PRIVATE_KEY"
+  deployer=$(cast wallet address $walletArgs)
+fi
+
+castSendHeader="cast send --rpc-url $networkURL$API_KEY $walletArgs"
+
+echo "${green}Whitelist SafeMultisigWithRecoveryModule${reset}"
+castArgs="$serviceRegistryAddress changeMultisigPermission(address,bool) $safeMultisigWithRecoveryModuleAddress true"
+echo $castArgs
+castCmd="$castSendHeader $castArgs"
+result=$($castCmd)
+echo "$result" | grep "status"
+
+echo "${green}Whitelist RecoveryModule${reset}"
+castArgs="$serviceRegistryAddress changeMultisigPermission(address,bool) $recoveryModuleAddress true"
+echo $castArgs
+castCmd="$castSendHeader $castArgs"
+result=$($castCmd)
+echo "$result" | grep "status"
--- a/scripts/deployment/l2/script_l2_02_change_managers_registries.sh
+++ b/scripts/deployment/l2/script_l2_02_change_managers_registries.sh
@@ -1,0 +1,87 @@
+#!/bin/bash
+
+# Check if $1 is provided
+if [ -z "$1" ]; then
+  echo "Usage: $0 <network>"
+  echo "Example: $0 eth_mainnet"
+  exit 1
+fi
+
+red=$(tput setaf 1)
+green=$(tput setaf 2)
+reset=$(tput sgr0)
+
+# Get globals file
+globals="$(dirname "$0")/globals_$1.json"
+if [ ! -f $globals ]; then
+  echo "${red}!!! $globals is not found${reset}"
+  exit 0
+fi
+
+# Read variables using jq
+useLedger=$(jq -r '.useLedger' $globals)
+derivationPath=$(jq -r '.derivationPath' $globals)
+chainId=$(jq -r '.chainId' $globals)
+networkURL=$(jq -r '.networkURL' $globals)
+
+componentRegistryAddress=$(jq -r '.componentRegistryAddress' $globals)
+agentRegistryAddress=$(jq -r '.agentRegistryAddress' $globals)
+registriesManagerAddress=$(jq -r '.registriesManagerAddress' $globals)
+serviceRegistryAddress=$(jq -r '.serviceRegistryAddress' $globals)
+serviceRegistryTokenUtilityAddress=$(jq -r '.serviceRegistryTokenUtilityAddress' $globals)
+serviceManagerProxyAddress=$(jq -r '.serviceManagerProxyAddress' $globals)
+
+# Check for Polygon keys only since on other networks those are not needed
+if [ $chainId == 137 ]; then
+  API_KEY=$ALCHEMY_API_KEY_MATIC
+  if [ "$API_KEY" == "" ]; then
+      echo "set ALCHEMY_API_KEY_MATIC env variable"
+      exit 0
+  fi
+elif [ $chainId == 80002 ]; then
+    API_KEY=$ALCHEMY_API_KEY_AMOY
+    if [ "$API_KEY" == "" ]; then
+        echo "set ALCHEMY_API_KEY_AMOY env variable"
+        exit 0
+    fi
+fi
+
+# Get deployer based on the ledger flag
+if [ "$useLedger" == "true" ]; then
+  walletArgs="-l --mnemonic-derivation-path $derivationPath"
+  deployer=$(cast wallet address $walletArgs)
+else
+  echo "Using PRIVATE_KEY: ${PRIVATE_KEY:0:6}..."
+  walletArgs="--private-key $PRIVATE_KEY"
+  deployer=$(cast wallet address $walletArgs)
+fi
+
+castSendHeader="cast send --rpc-url $networkURL$API_KEY $walletArgs"
+
+echo "${green}Change manager for ComponentRegistry${reset}"
+castArgs="$componentRegistryAddress changeManager(address) $registriesManagerAddress"
+echo $castArgs
+castCmd="$castSendHeader $castArgs"
+result=$($castCmd)
+echo "$result" | grep "status"
+
+echo "${green}Change manager for AgentRegistry${reset}"
+castArgs="$agentRegistryAddress changeManager(address) $registriesManagerAddress"
+echo $castArgs
+castCmd="$castSendHeader $castArgs"
+result=$($castCmd)
+echo "$result" | grep "status"
+
+echo "${green}Change manager for ServiceRegistry${reset}"
+castArgs="$serviceRegistryAddress changeManager(address) $serviceManagerProxyAddress"
+echo $castArgs
+castCmd="$castSendHeader $castArgs"
+result=$($castCmd)
+echo "$result" | grep "status"
+
+echo "${green}Change manager for ServiceRegistryTokenUtility${reset}"
+castArgs="$serviceRegistryTokenUtilityAddress changeManager(address) $serviceManagerProxyAddress"
+echo $castArgs
+castCmd="$castSendHeader $castArgs"
+result=$($castCmd)
+echo "$result" | grep "status"
--- a/scripts/deployment/l2/script_l2_03_service_manager_change_operator_whitelist.sh
+++ b/scripts/deployment/l2/script_l2_03_service_manager_change_operator_whitelist.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+
+# Check if $1 is provided
+if [ -z "$1" ]; then
+  echo "Usage: $0 <network>"
+  echo "Example: $0 eth_mainnet"
+  exit 1
+fi
+
+red=$(tput setaf 1)
+green=$(tput setaf 2)
+reset=$(tput sgr0)
+
+# Get globals file
+globals="$(dirname "$0")/globals_$1.json"
+if [ ! -f $globals ]; then
+  echo "${red}!!! $globals is not found${reset}"
+  exit 0
+fi
+
+# Read variables using jq
+useLedger=$(jq -r '.useLedger' $globals)
+derivationPath=$(jq -r '.derivationPath' $globals)
+chainId=$(jq -r '.chainId' $globals)
+networkURL=$(jq -r '.networkURL' $globals)
+
+serviceManagerProxyAddress=$(jq -r '.serviceManagerProxyAddress' $globals)
+operatorWhitelistAddress=$(jq -r '.operatorWhitelistAddress' $globals)
+
+# Check for Polygon keys only since on other networks those are not needed
+if [ $chainId == 137 ]; then
+  API_KEY=$ALCHEMY_API_KEY_MATIC
+  if [ "$API_KEY" == "" ]; then
+      echo "set ALCHEMY_API_KEY_MATIC env variable"
+      exit 0
+  fi
+elif [ $chainId == 80002 ]; then
+    API_KEY=$ALCHEMY_API_KEY_AMOY
+    if [ "$API_KEY" == "" ]; then
+        echo "set ALCHEMY_API_KEY_AMOY env variable"
+        exit 0
+    fi
+fi
+
+# Get deployer based on the ledger flag
+if [ "$useLedger" == "true" ]; then
+  walletArgs="-l --mnemonic-derivation-path $derivationPath"
+  deployer=$(cast wallet address $walletArgs)
+else
+  echo "Using PRIVATE_KEY: ${PRIVATE_KEY:0:6}..."
+  walletArgs="--private-key $PRIVATE_KEY"
+  deployer=$(cast wallet address $walletArgs)
+fi
+
+castSendHeader="cast send --rpc-url $networkURL$API_KEY $walletArgs"
+
+echo "${green}Change ServiceManager implementation in its proxy${reset}"
+castArgs="$serviceManagerProxyAddress setOperatorWhitelist(address) $operatorWhitelistAddress"
+echo $castArgs
+castCmd="$castSendHeader $castArgs"
+result=$($castCmd)
+echo "$result" | grep "status"
--- a/scripts/deployment/l2/script_l2_04_service_manager_change_implementation.sh
+++ b/scripts/deployment/l2/script_l2_04_service_manager_change_implementation.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+
+# Check if $1 is provided
+if [ -z "$1" ]; then
+  echo "Usage: $0 <network>"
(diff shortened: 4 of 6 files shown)


contracts/IdentityRegistryBridger.sol
```diff
@@ -0,0 +1,328 @@
...
+        // TODO Work on this block to finalize
+        // Assemble OLAS specific metadata entry
+        IIdentityRegistry.MetadataEntry[] memory metadataEntries = new IIdentityRegistry.MetadataEntry[](1);
+        metadataEntries[0] = IIdentityRegistry.MetadataEntry({
+            key: "Ecosystem",
+            value: abi.encode("OLAS V1")
+        });
+
+        // Create new agent Id
+        agentId = IIdentityRegistry(identityRegistry).register(tokenUri, metadataEntries);
```
kupermind: Call to the actual 8004 `IdentityRegistry`

contracts/ServiceManagerToken.sol
```diff
@@ -23,29 +31,47 @@ interface IToken {
...
-    /// @dev ServiceRegistryTokenUtility constructor.
-    /// @param _serviceRegistry Service Registry contract address.
-    /// @param _serviceRegistryTokenUtility Service Registry Token Utility contract address.
-    constructor(address _serviceRegistry, address _serviceRegistryTokenUtility, address _operatorWhitelist)
-        OperatorSignedHashes("Service Manager Token", "1.1.1")
+    /// @dev ServiceManager constructor.
+    /// @param _serviceRegistry Service Registry address.
+    /// @param _serviceRegistryTokenUtility Service Registry Token Utility address.
+    constructor(address _serviceRegistry, address _serviceRegistryTokenUtility)
+        OperatorSignedHashes(NAME, VERSION)
```
kupermind: Just a small optimization - don't hardcode domain name and version in the constructor, but rather have it as constants.

contracts/deprecated/ServiceManager.sol
```diff
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.15;
 
-import "./GenericManager.sol";
-import "./interfaces/IService.sol";
+import "../GenericManager.sol";
+import "../interfaces/IService.sol";
```
kupermind: Let's remove this contract completely as it's nowhere used anymore.

contracts/ServiceManagerToken.sol
```diff
@@ -23,29 +31,47 @@ interface IToken {
     function ownerOf(uint256 tokenId) external view returns (address);
 }
 
+/// @dev The contract is already initialized.
+error AlreadyInitialized();
```
mariapiamo: ?

contracts/ServiceManagerToken.sol
```diff
@@ -23,29 +31,47 @@ interface IToken {
...
+
 /// @title Service Manager - Periphery smart contract for managing services with custom ERC20 tokens or ETH
 /// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
-/// @author AL
+/// @author Andrey Lebedev - <andrey.lebedev@valory.xyz>
+/// @author Mariapia Moscatiello - <mariapia.moscatiello@valory.xyz>
 contract ServiceManagerToken is GenericManager, OperatorSignedHashes {
     event OperatorWhitelistUpdated(address indexed operatorWhitelist);
+    event IdentityRegistryBridgerUpdated(address indexed identityRegistryBridger);
+    event ImplementationUpdated(address indexed implementation);
```
mariapiamo: ?

contracts/ServiceManagerToken.sol
```diff
@@ -70,6 +114,39 @@ contract ServiceManagerToken is GenericManager, OperatorSignedHashes {
...
 
+    /// @dev Changes Identity Registry Bridger contract address.
+    /// @param newIdentityRegistryBridger New Identity Registry Bridger contract address.
+    function changeIdentityRegistryBridger(address newIdentityRegistryBridger) external {
+        // Check for the contract ownership
+        if (msg.sender != owner) {
+            revert OwnerOnly(msg.sender, owner);
+        }
+
+        identityRegistryBridger = newIdentityRegistryBridger;
```
mariapiamo: Come back to this to see what happens for zero address

contracts/ServiceManagerToken.sol
```diff
@@ -249,8 +337,15 @@ contract ServiceManagerToken is GenericManager, OperatorSignedHashes {
...
     {
+        // TODO Check if multisig was updated or not?
+        // Create or update multisig instance
         multisig = IService(serviceRegistry).deploy(msg.sender, serviceId, multisigImplementation, data);
+
+        // Register or update corresponding 8004 agent Id
+        (uint256 agentId, bool registered) = IIdentityRegistryBridger(identityRegistryBridger).register(serviceId);
+
         emit CreateMultisig(multisig);
+        emit ServiceAgentRegisteredOrUpdated(serviceId, agentId, registered);
```
mariapiamo: Can we also add olas agent id? Thinking for subgraph that will simplify lot of things

contracts/IdentityRegistryBridger.sol
```diff
@@ -0,0 +1,328 @@
...
+
+        // Check for multisig change
+        address agentServiceMultisig = IERC721(identityRegistry).getApproved(agentId);
+        // Check if multisig address has changed
+        if (serviceMultisig != agentServiceMultisig) {
+            // Approve updated service multisig such that it becomes agentId operator
+            IERC721(identityRegistry).approve(serviceMultisig, agentId);
+        }
+
+        // TODO Shall we NOT emit anything if nothing has been changed?
```
mariapiamo: Yeah not sure, at the end the data are the same, and we do not need to update essentially nothing. Though, for symmetry, I would keep it

contracts/IdentityRegistryBridger.sol
```diff
@@ -0,0 +1,328 @@
...
+            _update(serviceId, agentId);
+        }
+
+        _locked = 1;
+    }
+
+    /// @dev Links service Ids with created 8004 agent Ids.
+    /// @param serviceIds Set of service Ids.
+    /// @return agentIds Set of 8004 agent Ids.
+    function link(uint256[] memory serviceIds) external returns (uint256[] memory agentIds) {
```
mariapiamo: We need to discuss whether we need to check some access here

contracts/ServiceManagerToken.sol
```diff
@@ -23,29 +31,47 @@ interface IToken {
...
     event CreateMultisig(address indexed multisig);
+    event ServiceAgentRegisteredOrUpdated(uint256 indexed serviceId, uint256 indexed agentId, bool registered);
+
+    // Name of a signing domain
+    string public constant NAME = "OLAS Service Manager";
+    // Version number
+    string public constant VERSION = "1.2.0";
+    // Service Manager proxy address slot
+    // keccak256("PROXY_SERVICE_MANAGER") = "0xe39e69948a448ce9239ad71b908b6c5b46225f86ffa735b25a8cd64080315855"
+    bytes32 public constant PROXY_SERVICE_MANAGER = 0xe39e69948a448ce9239ad71b908b6c5b46225f86ffa735b25a8cd64080315855;
```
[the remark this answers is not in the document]
DavidMinarsch: No imo it does not - please revert this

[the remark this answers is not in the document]
DavidMinarsch: Unnecessary added complexity

contracts/ServiceManagerToken.sol
```diff
@@ -23,29 +31,47 @@ interface IToken {
...
     address public immutable serviceRegistry;
     // Service Registry Token Utility address
     address public immutable serviceRegistryTokenUtility;
-    // A well-known representation of ETH as an address
-    address public constant ETH_TOKEN_ADDRESS = address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE);
-    // Bond wrapping constant
-    uint96 public constant BOND_WRAPPER = 1;
+
+    // 8004 Identity Registry Bridger address
+    address public identityRegistryBridger;
```
DavidMinarsch: Please keep diffs minimal

contracts/ServiceManagerToken.sol
```diff
@@ -70,6 +114,39 @@ contract ServiceManagerToken is GenericManager, OperatorSignedHashes {
         emit OperatorWhitelistUpdated(newOperatorWhitelist);
     }
 
+    /// @dev Changes Identity Registry Bridger contract address.
+    /// @param newIdentityRegistryBridger New Identity Registry Bridger contract address.
+    function changeIdentityRegistryBridger(address newIdentityRegistryBridger) external {
```
DavidMinarsch: No we don't need that imo. We simply deploy a new periphery contract. Keep it simple stupid

contracts/ServiceManagerToken.sol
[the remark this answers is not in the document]
DavidMinarsch: IMO there should be no event here. We will track this on the IdentityRegistryBridge instead. Keeps the logic cleaner here no?

Alternatively, if we keep it here, needs to be called  more intuitively

contracts/ServiceManagerToken.sol
```diff
@@ -249,8 +337,15 @@ contract ServiceManagerToken is GenericManager, OperatorSignedHashes {
         bytes memory data
     ) external returns (address multisig)
     {
+        // TODO Check if multisig was updated or not?
+        // Create or update multisig instance
         multisig = IService(serviceRegistry).deploy(msg.sender, serviceId, multisigImplementation, data);
+
+        // Register or update corresponding 8004 agent Id
+        (uint256 agentId, bool registered) = IIdentityRegistryBridger(identityRegistryBridger).register(serviceId);
```
DavidMinarsch: pass multisig as arg, we already have it here, don't need to load from storage again then

contracts/IdentityRegistryBridger.sol
```diff
@@ -0,0 +1,328 @@
...
+        // Check for zero address, although this must never happen
+        if (multisig == address(0)) {
+            revert ZeroAddress();
+        }
+
+        // TODO Work on this block to finalize
+        // Assemble OLAS specific metadata entry
+        IIdentityRegistry.MetadataEntry[] memory metadataEntries = new IIdentityRegistry.MetadataEntry[](1);
+        metadataEntries[0] = IIdentityRegistry.MetadataEntry({
+            key: "Ecosystem",
```
DavidMinarsch: Please check the spec if we use the key correctly

Also they told me in the chat that the spec sets a number of keys as MUST - check which ones

contracts/IdentityRegistryBridger.sol
```diff
@@ -0,0 +1,328 @@
...
+            IIdentityRegistry(identityRegistry).setAgentUri(agentId, serviceTokenUri);
+        }
+
+        // Check for multisig change
+        address agentServiceMultisig = IERC721(identityRegistry).getApproved(agentId);
+        // Check if multisig address has changed
+        if (serviceMultisig != agentServiceMultisig) {
+            // Approve updated service multisig such that it becomes agentId operator
+            IERC721(identityRegistry).approve(serviceMultisig, agentId);
+        }
```
DavidMinarsch: We should handle on service manager side, keeps this code much cleaner

contracts/IdentityRegistryBridger.sol
```diff
@@ -0,0 +1,328 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.30;
+
+import {ERC721TokenReceiver} from "../lib/solmate/src/tokens/ERC721.sol";
+
+interface IERC721 {
+    /// @dev Gives permission to `spender` to transfer `tokenId` token to another account.
+    function approve(address spender, uint256 id) external;
```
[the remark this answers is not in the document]
kupermind: Yes, no need for all, but let's see with the 8004 operator

contracts/ServiceManagerToken.sol
kupermind: OK, here is just a bit of moving vars for better visibility to comply with all of our current developments.

contracts/IdentityRegistryBridger.sol
```diff
@@ -0,0 +1,307 @@
...
+    // keccak256("PROXY_IDENTITY_REGISTRY_BRIDGER") = "0x03684189c8fb7a536ac4dbd4b7ad063c37db21bcd0f9c51fe45a4eb16359c165"
+    bytes32 public constant PROXY_IDENTITY_REGISTRY_BRIDGER = 0x03684189c8fb7a536ac4dbd4b7ad063c37db21bcd0f9c51fe45a4eb16359c165;
+
+    // Identity Registry 8004 address
+    address public immutable identityRegistry;
+    // Service Registry address
+    address public immutable serviceRegistry;
+    // TODO immutable or not?
+    // 8004 Operator address
+    address public immutable operator;
```
kupermind: Need to work on this one as well

contracts/IdentityRegistryBridger.sol
kupermind: How are we going to link existing OLAS services?

contracts/ServiceManagerToken.sol
```diff
@@ -15,46 +15,106 @@ interface IOperatorWhitelist {
...
 /// @title Service Manager - Periphery smart contract for managing services with custom ERC20 tokens or ETH
 /// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
-/// @author AL
-contract ServiceManagerToken is GenericManager, OperatorSignedHashes {
+/// @author Andrey Lebedev - <andrey.lebedev@valory.xyz>
+/// @author Mariapia Moscatiello - <mariapia.moscatiello@valory.xyz>
+contract ServiceManager is GenericManager, OperatorSignedHashes {
     event OperatorWhitelistUpdated(address indexed operatorWhitelist);
+    event IdentityRegistryBridgerUpdated(address indexed identityRegistryBridger);
+    event ImplementationUpdated(address indexed implementation);
```
DavidMinarsch: not needed here

contracts/IdentityRegistryBridger.sol
DavidMinarsch: should be updatable by owner imo

contracts/IdentityRegistryBridger.sol
```diff
@@ -0,0 +1,307 @@
...
+            revert ManagerOnly(msg.sender, manager);
+        }
+
+        // Get corresponding agent Id
+        uint256 agentId = mapServiceIdAgentIds[serviceId];
+
+        // Modify only if agent Id is already defined, otherwise skip
+        if (agentId > 0) {
+            // Update agent wallet metadata entry
+            IIdentityRegistry(identityRegistry).setMetadata(agentId, "agentWallet", abi.encode(multisig));
```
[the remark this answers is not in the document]
kupermind: Correct

contracts/IdentityRegistryBridger.sol
```diff
@@ -0,0 +1,307 @@
...
+            revert();
+        }
+
+        // TODO Work on this block to finalize: multisig etc
+        // Assemble OLAS specific metadata entry
+        IIdentityRegistry.MetadataEntry[] memory metadataEntries = new IIdentityRegistry.MetadataEntry[](2);
+        metadataEntries[0] = IIdentityRegistry.MetadataEntry({
+            key: "Ecosystem",
+            value: abi.encode("OLAS V1")
+        });
```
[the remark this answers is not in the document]
kupermind: I would define a set of constants with metadata strings on top of the contract, also will be very visible as getters.

contracts/ERC8004Operator.sol
```diff
@@ -0,0 +1,194 @@
...
+/// @title ERC8004 Operator - Smart contract for managing requests in 8004 Identity Registry ecosystem
+/// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
+/// @author Andrey Lebedev - <andrey.lebedev@valory.xyz>
+/// @author Mariapia Moscatiello - <mariapia.moscatiello@valory.xyz>
+contract ERC8004Operator {
+    event OwnerUpdated(address indexed owner);
+    event FeedbackAuthSubmitted(address indexed sender, uint256 indexed agentId, address indexed clientAddress,
+        uint256 indexLimit, uint256 expiry, bytes32 digest);
+    event ValidationRequestSubmitted(address indexed sender, uint256 indexed agentId, address indexed validatorAddress,
+        string requestUri, bytes32 requestHash);
```
DavidMinarsch: Why do we need these events? The underlying registries will emit them

contracts/ERC8004Operator.sol
```diff
@@ -0,0 +1,194 @@
...
+
+        // Record "signed" digest
+        mapSignedHashes[digest] = true;
+
+        emit FeedbackAuthSubmitted(msg.sender, agentId, clientAddress, indexLimit, expiry, digest);
+
+        _locked - 1;
+    }
+
+    /// @dev Agent validation request.
```
DavidMinarsch: same

contracts/ServiceManagerToken.sol
```diff
@@ -400,4 +501,50 @@ contract ServiceManagerToken is GenericManager, OperatorSignedHashes {
...
     }
+
+    /// @dev Links service Ids with registered 8004 agent Ids.
+    /// @param startServiceId Starting service Id.
+    /// @param numServices Number of services to link.
+    /// @return agentIds Set of 8004 agent Ids.
+    function linkServiceIdAgentIds(
+        uint256 startServiceId,
+        uint256 numServices
+    ) external returns (uint256[] memory agentIds)
```
DavidMinarsch: I don't like to pollute SM with this.

We need to discuss this on a call as the deployment routine matters for prevention of out-of-sync. We might have to pause SM for some time whilst we backfill all this data..

DavidMinarsch: I think it's safer if we compute startServiceId from the on-chain state. We can just check what's the highest serviceId we have already minted as agentId

contracts/ERC8004Operator.sol
```diff
@@ -0,0 +1,194 @@
...
+    event OwnerUpdated(address indexed owner);
+    event FeedbackAuthSubmitted(address indexed sender, uint256 indexed agentId, address indexed clientAddress,
+        uint256 indexLimit, uint256 expiry, bytes32 digest);
+    event ValidationRequestSubmitted(address indexed sender, uint256 indexed agentId, address indexed validatorAddress,
+        string requestUri, bytes32 requestHash);
+
+    // Version number
+    string public constant VERSION = "0.1.0";
+    // Agent wallet multisig metadata key
+    string public constant AGENT_WALLET_MULTISIG_METADATA_KEY = "agentWallet: {multisig}";
```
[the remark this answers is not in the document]
kupermind: Oh that's how I understood it :) :) :) it was that you just provided a template. I thought it was weird :). Will fix.

contracts/IdentityRegistryBridger.sol
```diff
@@ -0,0 +1,308 @@
...
+    function changeManager(address newManager) external {
+        // Check for the ownership
+        if (msg.sender != owner) {
+            revert OwnerOnly(msg.sender, owner);
+        }
+
+        // Check for the zero address
+        if (newManager == address(0)) {
+            revert ZeroAddress();
+        }
```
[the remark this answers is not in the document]
kupermind: Since we check for the `msg.sender` to be a SM, if we disable here - `register()` et al will always revert unless the manager address is in place.

Do we want to do the following check for access control?
`if (manager != address(0) || (msg.sender == manager))`

contracts/IdentityRegistryBridger.sol
```diff
@@ -0,0 +1,456 @@
...
+            agentIds[i] = mapServiceIdAgentIds[serviceId];
+            // Check if agent Id has been registered
+            if (agentIds[i] == 0) {
+                // Register corresponding 8004 agent Id
+                _register(serviceId, multisig, tokenUri);
+            } else {
+                uint256 agentId = agentIds[i];
+
+                // Check agentWallet metadata
+                bytes memory agentWallet =
```
DavidMinarsch: I think a simpler routine would be (given we know agentId exists):

-> check if multisig in mapMultisigAgentIds, if so, we know by design it's up-to-date
-> if not, update mapMultisigAgentIds and on-chain metadata of this agentId

contracts/ServiceManager.sol
```diff
@@ -1,68 +1,311 @@
...
+        serviceRegistryTokenUtility = _serviceRegistryTokenUtility;
+        identityRegistryBridger = _identityRegistryBridger;
+        operatorWhitelist = _operatorWhitelist;
+
         owner = msg.sender;
     }
 
+    /// @dev Sets the operator whitelist contract address.
+    /// @param newOperatorWhitelist New operator whitelist contract address.
+    function setOperatorWhitelist(address newOperatorWhitelist) external {
```
DavidMinarsch: where was this before?

kupermind: It's the original one

contracts/IdentityRegistryBridger.sol
```diff
@@ -0,0 +1,456 @@
...
+
+        emit StartLinkServiceIdUpdated(serviceId);
+    }
+
+    /// @dev Updates or links service Ids with registered 8004 agent Ids.
+    /// @param serviceIds Set of service Ids.
+    /// @return agentIds Corresponding set of 8004 agent Ids.
+    function updateOrLinkServiceIdAgentIds(uint256[] memory serviceIds) public returns (uint256[] memory agentIds) {
+        // Get number of service Ids
+        uint256 numServices = serviceIds.length;
```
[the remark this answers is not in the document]
kupermind: `updateOrLinkServiceIdAgentId(serviceId)` should work with any set of serviceIds, as this function will mostly be used after `ServiceManager` integration. The contiguous numbers array will most likely not be needed there.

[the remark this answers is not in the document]
kupermind: But yeah, let's make it in place without `serviceIds` allocation, such that it does not look stupid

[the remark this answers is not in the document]
kupermind: Addressed

mariapiamo (approved):

PR #241, opened 2025-11-14
refactor: ServiceManager to account for IdentityRegistryBridger

kupermind:
Extending OLAS agent registries to account for seamless 8004 bridging:
- Existing Olas AI agent registries, controlled via `ServiceManager` are extended to mint a corresponding `ERC-8004` instance for each Olas agent.
- `IdentityRegistryBridger.sol` acts as a bridge to ERC8004 IdentityRegistry, that state between both Olas and ERC-8004 registries stays synced.

### DIFF
--- a/docs/configuration.json
+++ b/docs/configuration.json
@@ -426,6 +426,16 @@
         "name": "ComplementaryServiceMetadata",
         "artifact": "abis/0.8.30/ComplementaryServiceMetadata.json",
         "address": "0x11949cBC85d8793B360029E26b18ae759708e28b"
+      },
+      {
+        "name": "IdentityRegistryBridger",
+        "artifact": "abis/0.8.30/IdentityRegistryBridger.json",
+        "address": "0x88943F63E29cd436B62cFfE332aD54De92AdCE98"
+      },
+      {
+        "name": "IdentityRegistryBridgerProxy",
+        "artifact": "abis/0.8.30/IdentityRegistryBridgerProxy.json",
+        "address": "0x7fc0ddf4DFB61CfA5519db2A5eE7B2Eb02De0140"
       }
     ]
   },
--- a/scripts/deployment/l2/globals_optimism_mainnet.json
+++ b/scripts/deployment/l2/globals_optimism_mainnet.json
@@ -39,5 +39,7 @@
   "serviceManagerProxyAddress": "0xA5C7FbCCFf28441b7d250412b0Fb87AA1c8b14AD",
   "hashCheckpointAddress": "0x4E82BC73BFB9647074aA71CBF467e66c1808C571",
   "complementaryServiceMetadataAddress": "0x11949cBC85d8793B360029E26b18ae759708e28b",
-  "agent8004BaseURI": "https://marketplace.olas.network/erc8004/op-mainnet/ai-agents/"
+  "agent8004BaseURI": "https://marketplace.olas.network/erc8004/op-mainnet/ai-agents/",
+  "identityRegistryBridgerAddress": "0x88943F63E29cd436B62cFfE332aD54De92AdCE98",
+  "identityRegistryBridgerProxyAddress": "0x7fc0ddf4DFB61CfA5519db2A5eE7B2Eb02De0140"
 }


contracts/ServiceManager.sol
```diff
@@ -234,6 +298,18 @@ contract ServiceManager is GenericManager, OperatorSignedHashes {
...
+        if (identityRegistryBridger != address(0)) {
+            // Get updated service token URI
+            string memory updatedTokenUri = IERC721(serviceRegistry).tokenURI(serviceId);
+
+            // Check if tokenUri has changed
+            if (keccak256(bytes(curServiceTokenUri)) != keccak256(bytes(updatedTokenUri))) {
+                // Updated tokenUri in 8004 Identity Registry, if required
+                IIdentityRegistryBridger(identityRegistryBridger).updateAgentUri(serviceId, updatedTokenUri);
+            }
+        }
```
DavidMinarsch: drop

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -181,27 +180,26 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
     /// @return agentId Corresponding 8004 agent Id.
     function _register(uint256 serviceId, address multisig, string memory tokenUri) internal returns (uint256 agentId)  {
```
DavidMinarsch: no need for tokenUri as arg

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -181,27 +180,26 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
...
-        });
-        metadataEntries[2] = IIdentityRegistry.MetadataEntry({
             key: SERVICE_REGISTRY_METADATA_KEY,
             value: abi.encode(serviceRegistry)
         });
-        metadataEntries[3] = IIdentityRegistry.MetadataEntry({
+        metadataEntries[2] = IIdentityRegistry.MetadataEntry({
             key: SERVICE_ID_METADATA_KEY,
             value: abi.encode(serviceId)
         });
```
DavidMinarsch: agentUri = baseUri + "/{serviceId}"

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -324,17 +323,20 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
...
         // Get corresponding 8004 agent Id
         agentId = mapServiceIdAgentIds[serviceId];
-        // Only allow one agentId per serviceId
-        if (agentId > 0) {
-            revert AgentIdAlreadyAssigned(agentId, serviceId);
-        }
+        // Only allow one agentId per serviceId: skip if agentId has its corresponding serviceId
+        if (agentId == 0) {
+            // Get service token URI
+            string memory tokenUri = IERC721(serviceRegistry).tokenURI(serviceId);
```
DavidMinarsch: drop

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -114,15 +111,17 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
     string public constant ECOSYSTEM_METADATA_KEY = "ecosystem";
     // Ecosystem metadata value
```
DavidMinarsch, line 86: Add setter (owner only) and constructor arg for baseURI

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -16,11 +12,11 @@ interface IIdentityRegistry {
...
     function register(string memory tokenUri, MetadataEntry[] memory metadata) external returns (uint256 agentId);
 
-    function setMetadata(uint256 agentId, string memory key, bytes memory value) external;
+    function getMetadata(uint256 agentId, string memory key) external view returns (bytes memory);
 
-    function setAgentUri(uint256 agentId, string calldata newUri) external;
+    function setAgentWallet(uint256 agentId, address newWallet, uint256 deadline, bytes calldata signature) external;
 
-    function getMetadata(uint256 agentId, string memory key) external view returns (bytes memory);
+    function unsetAgentWallet(uint256 agentId) external;
```
DavidMinarsch, line 19: Please double check we support all methods from IR - e.g.what about setMetadata?

We need to expose setMetadata via IRB so agent wallets can update their metadata properly

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -114,15 +99,18 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
     string public constant ECOSYSTEM_METADATA_KEY = "ecosystem";
     // Ecosystem metadata value
     bytes public constant ECOSYSTEM_METADATA_VALUE = abi.encode("OLAS V1");
```
DavidMinarsch:
```suggestion
    bytes public constant ECOSYSTEM_METADATA_VALUE = abi.encode("Olas");
```

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -456,23 +471,24 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
...
-            (,address multisig,,,,,) = IServiceRegistry(serviceRegistry).mapServices(serviceId);
-            // Skip services that were never deployed
-            if (multisig == address(0)) {
-                continue;
-            }
-
-            // Get service token URI
-            string memory tokenUri = IERC721(serviceRegistry).tokenURI(serviceId);
 
             // Get corresponding 8004 agent Id
```
DavidMinarsch, line 468: Let's make sure linkServiceIdAgentIds can only be called when not all services are already linked?

contracts/8004/IdentityRegistryBridger.sol
kupermind: Shall we restrict which metadata they can update? For example, they should not change serviceRegistry and serviceId metadata? Basically all default ones?

DavidMinarsch: yes sounds good, restrict editing of the keys we set

scripts/deployment/globals_sepolia.json
```diff
@@ -41,7 +41,7 @@
   "stakingFactoryAddress": "",
   "recoveryModuleAddress": "0x0a40C5ff1EB345E92DaAB5981Fc05219cCe4463E",
   "safeMultisigWithRecoveryModuleAddress": "0x164e1CA068afeF66EFbB9cA19d904c44E8386fd9",
-  "agent8004BaseURI": "https://gateway.autonolas.tech/erc8004/ethereum/service/",
+  "agent8004BaseURI": "https://marketplace.olas.network/erc8004/ethereum/ai-agents/",
```
rajat2502, line 44: it this expected for sepolia (i guess it is as per the previous code, just asking for my understanding)

kupermind: Good catch there, I was thinking about it as well for the moment :) however on sepolia we try to mimic what happens on mainnet. Those contracts there are test wallet owned, and thus we just continue simulate the mainnet approach.

rajat2502: Got it, thanks for explaining.

PR #247, opened 2025-12-15
feat: PolySafe creator with protocol recovery module

kupermind:
- PolySafe creator with protocol recovery module.

### DIFF
--- a/test/ServiceStakingGasLimit.t.sol
+++ b/test/ServiceStakingGasLimit.t.sol
@@ -1,0 +1,288 @@
+pragma solidity ^0.8.30;
+
+import {IService} from "../contracts/interfaces/IService.sol";
+import "@gnosis.pm/safe-contracts/contracts/GnosisSafe.sol";
+import {GnosisSafeProxy} from "@gnosis.pm/safe-contracts/contracts/proxies/GnosisSafeProxy.sol";
+import {GnosisSafeProxyFactory} from "@gnosis.pm/safe-contracts/contracts/proxies/GnosisSafeProxyFactory.sol";
+import {Test, console} from "forge-std/Test.sol";
+import {Utils} from "./utils/Utils.sol";
+import {ERC20Token} from "../contracts/test/ERC20Token.sol";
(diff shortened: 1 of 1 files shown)


contracts/multisigs/PolySafeCreatorWithRecoveryModule.sol
```diff
@@ -0,0 +1,126 @@
...
+/// @dev Provided incorrect multisig owner.
+/// @param provided Provided owner address.
+error WrongOwner(address provided);
+
+/// @title PolySafeCreatorWithRecoveryModule - Smart contract for Poly Safe multisig implementation with Recovery Module
+/// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
+/// @author Andrey Lebedev - <andrey.lebedev@valory.xyz>
+/// @author Mariapia Moscatiello - <mariapia.moscatiello@valory.xyz>
+contract PolySafeCreatorWithRecoveryModule {
+    event MultisigCreated(address indexed multisig, address indexed owner);
```
mariapiamo: I think we should keep just multisig to be compatible with subgraph. But worth to check that with Axat

contracts/multisigs/PolySafeCreatorWithRecoveryModule.sol
```diff
@@ -0,0 +1,207 @@
...
+
+        // Enable module payload
+        bytes memory execData = abi.encodeCall(ISafe.enableModule, (recoveryModule));
+
+        // Enable Recovery Module
+        // to = multisig, value = 0, operation = Call, all payment related = 0
+        ISafe(multisig)
+            .execTransaction(
+                multisig, 0, execData, ISafe.Operation.Call, 0, 0, 0, address(0), payable(address(0)), enableModuleSig
+            );
```
kupermind: `enableModule()` encoding and call. The signature must be provided along with polySafe creation signature

mariapiamo:
As agreed today, we will have a test  trying to recover when the owner is a gnosis safe multisig v 1.3.0

77ph (approved):

mariapiamo (approved):
we will submit updated vulnerabilities list in a following pr

PR #259, opened 2026-01-13
8004 extension2

kupermind:
- 8004 v1.0 compatibility.


contracts/ServiceManager.sol
```diff
@@ -19,9 +19,8 @@ interface IIdentityRegistryBridger {
     /// @dev Registers 8004 agent Id corresponding to service Id.
```
DavidMinarsch, line 19: Throughout we simplify to let Bridger handle existence checks. That's it right?

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -308,9 +305,8 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
     /// @dev Registers 8004 agent Id corresponding to service Id.
     /// @param serviceId Service Id.
     /// @param multisig Service multisig.
-    /// @param tokenUri Service tokenUri.
     /// @return agentId Corresponding 8004 agent Id.
-    function register(uint256 serviceId, address multisig, string memory tokenUri) external returns (uint256 agentId) {
+    function register(uint256 serviceId, address multisig) external returns (uint256 agentId) {
```
DavidMinarsch, line 312: simplified - lgtm

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -181,20 +180,16 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
...
     function _register(uint256 serviceId, address multisig, string memory tokenUri) internal returns (uint256 agentId)  {
         // Assemble OLAS specific metadata entry
-        IIdentityRegistry.MetadataEntry[] memory metadataEntries = new IIdentityRegistry.MetadataEntry[](4);
+        IIdentityRegistry.MetadataEntry[] memory metadataEntries = new IIdentityRegistry.MetadataEntry[](3);
         metadataEntries[0] = IIdentityRegistry.MetadataEntry({
             key: ECOSYSTEM_METADATA_KEY,
             value: ECOSYSTEM_METADATA_VALUE
         });
         metadataEntries[1] = IIdentityRegistry.MetadataEntry({
-            key: AGENT_WALLET_METADATA_KEY,
```
DavidMinarsch, line 190: We remove because Erc8004 registry does it for us automatically, right?

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -181,20 +180,16 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
...
-        });
-        metadataEntries[2] = IIdentityRegistry.MetadataEntry({
             key: SERVICE_REGISTRY_METADATA_KEY,
             value: abi.encode(serviceRegistry)
         });
-        metadataEntries[3] = IIdentityRegistry.MetadataEntry({
+        metadataEntries[2] = IIdentityRegistry.MetadataEntry({
             key: SERVICE_ID_METADATA_KEY,
             value: abi.encode(serviceId)
         });
```
DavidMinarsch, line 195: After
```
agentId = IIdentityRegistry(identityRegistry).register(tokenUri, metadataEntries);
```
We need to call `unsetAgentWallet` on IdentityRegistry to ensure the bridger isn't now the agent wallet

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -387,7 +388,56 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
...
+        address oldMultisig;
+        if (metadata.length == ADDRESS_BYTES_SIZE) {
+            // solhint-disable-next-line avoid-low-level-calls
+            assembly {
+                oldMultisig := mload(add(metadata, ADDRESS_BYTES_SIZE))
+            }
+        }
+
+        // Set agent wallet on behalf of agent
+        IIdentityRegistry(identityRegistry).setAgentWallet(agentId, msg.sender, deadline, "");
```
DavidMinarsch, line 434: And this will pass because?

contracts/8004/IdentityRegistryBridger.sol
kupermind: Correct, during the `setAgentWallet()` function execution.

contracts/8004/IdentityRegistryBridger.sol
kupermind: It will pass if the corresponding signature is in the `msg.sender`. Otherwise it reverts.

DavidMinarsch: We can drop this now as per chat on Slack

contracts/multisigs/SafeMultisigWithRecoveryModule8004.sol
```diff
@@ -0,0 +1,285 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.30;
```
DavidMinarsch: This file can be deleted as we won't take this approach

contracts/test/CompatibilityFallbackHandler.sol
```diff
@@ -0,0 +1,159 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity >=0.7.0 <0.9.0;
```
DavidMinarsch, line 2: Same, won't take this approach

PR #263, opened 2026-01-21
refactor: adjust to latest 8004 code workflow

kupermind:
- Adjust to latest 8004 code workflow.

### DIFF
--- a/contracts/8004/IdentityRegistryBridger.sol
+++ b/contracts/8004/IdentityRegistryBridger.sol
@@ -9,12 +9,7 @@ interface IERC721 {
 }
 
 interface IIdentityRegistry {
-    struct MetadataEntry {
-        string key;
-        bytes value;
-    }
-
-    function register(string memory tokenUri, MetadataEntry[] memory metadata) external returns (uint256 agentId);
+    function register() external returns (uint256 agentId);
 
     function setMetadata(uint256 agentId, string memory key, bytes memory value) external;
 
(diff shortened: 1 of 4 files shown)


contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -175,34 +180,20 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
...
-        metadataEntries[1] = IIdentityRegistry.MetadataEntry({
-            key: SERVICE_REGISTRY_METADATA_KEY,
-            value: abi.encode(serviceRegistry)
-        });
-        metadataEntries[2] = IIdentityRegistry.MetadataEntry({
-            key: SERVICE_ID_METADATA_KEY,
-            value: abi.encode(serviceId)
-        });
-
+    function _register(uint256 serviceId) internal returns (uint256 agentId)  {
```
kupermind, line 184: Register is called when the service is created, or when previously deployed services get their 8004 agents. No `agentURI` is set at this stage.

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -222,13 +213,11 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
     /// @param oldMultisig Old multisig address.
     /// @param newMultisig New multisig address.
     function _updateAgentWallet(uint256 serviceId, uint256 agentId, address oldMultisig, address newMultisig) internal {
-        // Update agent wallet and metadata entry
-        IIdentityRegistry(identityRegistry).setAgentWallet(agentId, newMultisig, block.timestamp, "");
```
kupermind, line 226: `setAgentWallet()` of `IdentityRegistry` contract can only be called by the serviceRegistry deploy() function, as the signature is being added, or via the external `setAgentWallet()` function of this contract.

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -136,6 +139,8 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
     address public owner;
     // Manager address
     address public manager;
+    // Service Id <-> agent Id linker
+    address public linker;
```
kupermind, line 143: Linker is a specific account that takes care of updating services to 8004 compatible agents.

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -322,11 +327,8 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
         agentId = mapServiceIdAgentIds[serviceId];
         // Only allow one agentId per serviceId: skip if agentId has its corresponding serviceId
         if (agentId == 0) {
-            // Get service token URI
-            string memory tokenUri = IERC721(serviceRegistry).tokenURI(serviceId);
-
             // Register new agentId
-            agentId = _register(serviceId, multisig, tokenUri);
+            agentId = _register(serviceId);
```
kupermind, line 331: We don't pass multisig here as by default `register()` is called during service creation when supposedly multisig is unknown.

PR #266, opened 2026-01-23
refactor: address latest comments

kupermind:
- Address latest comments.

### DIFF
--- a/contracts/8004/IdentityRegistryBridger.sol
+++ b/contracts/8004/IdentityRegistryBridger.sol
@@ -69,10 +69,6 @@ error ReentrancyGuard();
 /// @param metadataKey Metadata key.
 error WrongMetadataKey(string metadataKey);
 
-/// @dev Provided wrong service Id.
-/// @param serviceId Service Id.
-error WrongServiceId(uint256 serviceId);
-
 /// @title Identity Registry Bridger - Smart contract for bridging OLAS AI Agent Registry with 8004 Identity Registry
 /// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
 /// @author Andrey Lebedev - <andrey.lebedev@valory.xyz>
--- a/contracts/8004/IdentityRegistryBridger.sol
+++ b/contracts/8004/IdentityRegistryBridger.sol
@@ -12,11 +12,13 @@ interface IIdentityRegistry {
 
     function register(string memory tokenUri, MetadataEntry[] memory metadata) external returns (uint256 agentId);
 
-    function getMetadata(uint256 agentId, string memory key) external view returns (bytes memory);
+    function setMetadata(uint256 agentId, string memory metadataKey, bytes memory metadataValue) external;
 
     function setAgentWallet(uint256 agentId, address newWallet, uint256 deadline, bytes calldata signature) external;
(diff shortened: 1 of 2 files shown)


contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -389,14 +386,9 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
...
+        // TODO Is this also correct when metadata == "0x"?
         // Decode multisig value
-        address oldMultisig;
-        if (metadata.length == ADDRESS_BYTES_SIZE) {
-            // solhint-disable-next-line avoid-low-level-calls
-            assembly {
-                oldMultisig := mload(add(metadata, ADDRESS_BYTES_SIZE))
-            }
-        }
+        address oldMultisig = address(bytes20(metadata));
```
kupermind, line 391: I believe in recent versions of compiler one could not cast like that into address, so no assembly is needed

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -389,14 +386,9 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
 
         // Get old multisig address
         metadata = IIdentityRegistry(identityRegistry).getMetadata(agentId, AGENT_WALLET_METADATA_KEY);
+        // TODO Is this also correct when metadata == "0x"?
```
kupermind, line 389: For @77ph

contracts/test/MockIdentityRegistry.sol
```diff
@@ -5310,10 +5305,8 @@ contract MockIdentityRegistry is
     /// @custom:storage-location erc7201:erc8004.identity.registry
     struct IdentityRegistryStorage {
         uint256 _lastId;
-        // agentId => metadataKey => metadataValue
+        // agentId => metadataKey => metadataValue (includes "agentWallet")
```
kupermind, line 5308: Updated up to the latest 8004 version that is going to be deployed

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -406,6 +398,40 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
...
+            keccak256(bytes(metadataKey)) == keccak256(bytes(SERVICE_REGISTRY_METADATA_KEY)) ||
+            keccak256(bytes(metadataKey)) == keccak256(bytes(SERVICE_ID_METADATA_KEY))
+        ) {
+            revert WrongMetadataKey(metadataKey);
+        }
+
+        // Set metadata
+        IIdentityRegistry(identityRegistry).setMetadata(agentId, metadataKey, metadataValue);
+
+        emit MetadataSet(agentId, metadataKey, metadataValue);
```
DavidMinarsch, line 430: That just repeats the event on IdentityRegistry. Not sure it's worth it. But ok

DavidMinarsch (approved):

contracts/8004/IdentityRegistryBridger.sol
kupermind: We can rename it. In general, we try to emit events in all of our contract functions.

PR #275, opened 2026-02-02
refactor: fix agent wallet set event

kupermind:
- Fix agent wallet set event.


contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -350,23 +349,10 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
...
-        // Get zero address old multisig by default, if metadata is empty
-        address oldMultisig;
-
-        // Check metadata length and decode accordingly
-        if (metadata.length == 20) {
-            oldMultisig= address(bytes20(metadata));
-        } else if (metadata.length == 32) {
-            oldMultisig= abi.decode(metadata, (address));
-        }
-
```
kupermind, line 365: No need to fetch old multisig because if this function is called, the old multisig is always zero address since:
- If the service was just deployed - previous agent wallet is zero address due to unset one when the agent is registered;
- If the multisig was updated during service re-deployment - it calls `updateAgentWallet()` function in the same transaction, also unsetting old multisig.

Thus, `oldMultisig` value will always be zero address at this state.

77ph (approved):

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -350,23 +349,10 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
...
         // Set agent wallet on behalf of agent
         IIdentityRegistry(identityRegistry).setAgentWallet(agentId, msg.sender, deadline, signature);
 
-        emit AgentMultisigUpdated(serviceId, agentId, oldMultisig, msg.sender);
+        emit AgentWalletSet(serviceId, agentId, msg.sender);
```
DavidMinarsch, line 357: Do we know if on Ethereum any agent has already called setAgentWallet?

mariapiamo (approved):

contracts/8004/IdentityRegistryBridger.sol
DavidMinarsch: I don't follow the argument.

The point is we don't need the oldMultisig value below so we don't need to retrieve it. Independently of whether you're right or wrong..

contracts/8004/IdentityRegistryBridger.sol
kupermind: I believe not, as the MW is not yet updated. Also, the number of service Id-s has not changed since the linking was performed.

contracts/8004/IdentityRegistryBridger.sol
kupermind: @DavidMinarsch we don't retrieve the `oldMultisig` value because it will always be zero address, and thus it's meaningless.

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -75,7 +75,8 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
     event AgentMultisigUpdated(
         uint256 indexed serviceId, uint256 indexed agentId, address oldMultisig, address indexed newMultisig
     );
-    event MetadataSet(uint256 indexed agentId, string metadataKey, bytes metadataValue);
+    event AgentWalletSet(uint256 indexed serviceId, uint256 indexed agentId, address indexed multisig);
+    event MetadataSet(uint256 indexed serviceId, uint256 indexed agentId, string metadataKey, bytes metadataValue);
```
kupermind, line 79: For symmetry - adding `serviceId` as well.

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -150,6 +149,28 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
         return string(abi.encodePacked(baseURI, LibString.toString(serviceId)));
     }
 
+    /// @dev Gets agent Id and service Id by multisig agent wallet.
+    /// @param multisig Multisig address.
+    /// @return agentId 8004 agent Id.
+    /// @return serviceId Service Id.
+    function _getAgentServiceIds(address multisig) internal view returns (uint256 agentId, uint256 serviceId) {
```
kupermind, line 149: Just a simple copy-paste code reduction optimization - this is fetched in 3 places now.

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -334,39 +355,34 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
...
-        // Set agent wallet on behalf of agent
-        IIdentityRegistry(identityRegistry).setAgentWallet(agentId, msg.sender, deadline, signature);
+        // Get agent Id and service Id
+        (uint256 agentId, uint256 serviceId) = _getAgentServiceIds(msg.sender);
+
+        // Unset agent wallet on behalf of agent
+        IIdentityRegistry(identityRegistry).unsetAgentWallet(agentId);
 
-        emit AgentMultisigUpdated(serviceId, agentId, oldMultisig, msg.sender);
+        emit AgentWalletSet(serviceId, agentId, msg.sender);
```
DavidMinarsch: Needs to emit zero address
```suggestion
        emit AgentWalletSet(serviceId, agentId, 0x0);
```

DavidMinarsch (approved):

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -95,8 +96,6 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
     string public constant SERVICE_REGISTRY_METADATA_KEY = "serviceRegistry";
     // Service Id metadata key
     string public constant SERVICE_ID_METADATA_KEY = "serviceId";
-    // Agent wallet multisig metadata key
-    string public constant AGENT_WALLET_METADATA_KEY = "agentWallet";
```
mariapiamo, line 99: we remove this since we already are able to get this info from getAgentWallet? In that case, should we also wrap that getter?

contracts/8004/IdentityRegistryBridger.sol
```diff
@@ -75,7 +75,8 @@ contract IdentityRegistryBridger is ERC721TokenReceiver {
     event AgentMultisigUpdated(
         uint256 indexed serviceId, uint256 indexed agentId, address oldMultisig, address indexed newMultisig
     );
-    event MetadataSet(uint256 indexed agentId, string metadataKey, bytes metadataValue);
+    event AgentWalletSet(uint256 indexed serviceId, uint256 indexed agentId, address indexed multisig);
+    event MetadataSet(uint256 indexed serviceId, uint256 indexed agentId, string metadataKey, bytes metadataValue);
     event ValidationRequestSubmitted(
```
mariapiamo: should we also add serviceId here for symmetry?

contracts/8004/IdentityRegistryBridger.sol
DavidMinarsch: why would we wrap a getter?

mariapiamo: yeah, no need. We do not use it into the code indeed

contracts/8004/IdentityRegistryBridger.sol
kupermind: Let's remove it for now

PR #290, opened 2026-04-29
doc: internal17 corrections — verification proofs, L-11 reclassification, two-column status matrix

kupermind:
## Summary

Audit-internal17 corrections in response to colleague review feedback, plus a structural change to the verification matrix (modelled on `autonolas-governance/audits/internal19`):

- **§4.2 H-06 proof was citing the wrong entry point** (`registerAgentsWithSignature` lines instead of the `create` path). Replaced with the actual create-path guard: `ServiceManager.create:203–206/255` and `ServiceRegistry.create:226–229/274`, with `_safeMint` at line 270 inside the lock window.
- **§4.2 incorrectly stated `ServiceRegistry.registerAgents` is "lock-guarded analogously".** It isn't — the inline `_locked` block exists only on `create`, `deploy`, `terminate`, `unbond`, `drain`. `registerAgents` / `update` / `activateRegistration` are mitigated by the role-gated `ServiceManager` periphery, which **is** lock-guarded. Caveat added.
- **§4.8 L-11 reclassified** from "fixed at ServiceManager layer" to documented trade-off. The `multisig != address(0)` check in `ServiceManager.deploy:434–439` is *a* guard but not the L-11 guard — in `PolySafeCreatorWithRecoveryModule.create`, the proxy is fully deployed and codehash-checked before the `execTransaction(enableModule)` call, so the returned `multisig` is non-zero regardless of whether the module-enabling call silently failed. What protects the deployed system today is Safe v1.3.0's revert semantics under `safeTxGas = 0, gasPrice = 0`, not an explicit success check. Recommendation: add an explicit on-chain bool / post-condition check for defence-in-depth.
- **§4.9 L-12 msg.value description corrected**: token-secured branch forwards `length * BOND_WRAPPER` (excess trapped); native branch forwards `msg.value` (revert on mismatch); `ServiceRegistry.registerAgents:427` checks `msg.value != totalBond` (sum of stored bonds, only equal to `length * BOND_WRAPPER` for token-secured services).
- **Two-column status vocabulary** added in §0 — `Code status` (`✅ Fixed in main`, `📝 Documented`, etc.) and `Deployment status` (`🟢 Live on-chain`, `🟡 Pending redeploy`, `⚪ Code fix only — never deployed`, ...). Top-of-doc table, §4 matrix, §4.12 summary, §6.1 findings, §8 conclusion all re-tabulated.
- **H-05 / H-10 deployment-state honest framing**: post-fix `StakingToken` / `StakingNativeToken` source on HEAD `d60f7ef5` has **never been deployed**. Plan: register as a new allowlisted implementation in `StakingVerifier` alongside the existing one (not replacing it); pre-existing `StakingProxy` clones remain on their original implementation by design (constructor-pinned, §5.5).
- **Vulnerabilities_list_registries.md**: item #13 (L-11) rewritten for accuracy; new item #22 added documenting `ServiceRegistry.registerAgents` / `update` / `activateRegistration` missing inline reentrancy guard (role-based mitigation).

## Test plan

- [ ] Re-read the updated `audits/internal17/README.md` end-to-end and confirm the two-column matrix is internally consistent.
- [ ] Verify `docs/Vulnerabilities_list_registries.md` item #13 (L-11) and the new item #22 read correctly with the surrounding entries.
- [ ] Decide whether to:
  - (a) fix L-11 in code by checking the `execTransaction` bool (or asserting the recovery module is enabled post-call) in `PolySafeCreatorWithRecoveryModule.create`, in which case L-11 promotes back to `✅ Fixed in origin/main`;
  - (b) keep L-11 as a documented trade-off and ship the audit as-is.
- [ ] Plan the deployment of the post-fix `StakingToken` / `StakingNativeToken` as new allowlisted implementations in `StakingVerifier` on the chains where it matters (defence-in-depth, not blocking, per §3.1 / §4.1).


mariapiamo (commented):
Rechecked the latest PR head: `b68ecb951b12d6e9bbf8be42f2c0b09f742bb56f`.

The main L-11 / H-06 / `msg.value` corrections in `audits/internal17/README.md` are much improved, and the Safe v1.3.0 `GS013` rationale is supported by the installed Safe source. I still see a few doc-consistency issues to fix before this is clean:

1. `docs/Vulnerabilities_list_registries.md` item #15 still has the stale L-12 wording. It says the native path checks `agentInstances.length * BOND_WRAPPER` and the signature path always forwards that amount. The corrected logic is the one now described in the audit README: token-secured services forward `length * BOND_WRAPPER`; native services forward `msg.value`; `ServiceRegistry.registerAgents` checks `msg.value != totalBond`.

2. `audits/internal17/README.md` section 7 still says the vulnerabilities list tracks 21 items and the table stops at #21, while this PR adds item #22 and the conclusion says 22 entries. The methodology table also still says 21 entries. Item #22 should either be included in the section 7 verification table or the section should explicitly state that #22 is handled separately.

3. The §3.1 derivative-lock test coverage paragraph is still stale: it says `test/StakingDerivativeReentrancy.t.sol` is +260 LOC and lists six tests. At this PR head the file is 399 lines and contains four tests: `test_Reentrancy_NestedReceive_RevertsWithGuard`, `test_Reentrancy_NestedDeposit_RevertsWithGuard`, `test_Receive_HonestDepositSucceeds`, and `test_Deposit_HonestDepositSucceeds`. The nearby statement that the outer `_withdraw` necessarily reverts is also too broad; the actual nested receive test swallows the inner revert and the outer claim completes.

4. New in `audits/internal17/FINAL_REVIEW.md`: H-05 says there are “5 nested-entry tests” in `test/StakingDerivativeReentrancy.t.sol`; same correction as above applies. The metadata also says composite tip `c59e201`, but the current PR head is `b68ecb9`.

Evidence from my latest pass is in `/tmp/autonolas_pr290_review/summary.md` locally.

mariapiamo (approved):

77ph (approved):

PR #296, opened 2026-05-22
feat(scripts): ServiceManagerProxy & IdentityRegistryBridgerProxy change-owner scripts

kupermind:
Adds idempotent ownership-transfer scripts for `ServiceManagerProxy` and `IdentityRegistryBridgerProxy`, which an on-chain owner audit found are still owned by the deployer EOA rather than governance on every chain.

## What's added

- `scripts/deployment/script_l1_07_service_manager_proxy_change_owner.sh` — run with `mainnet`
- `scripts/deployment/script_l1_08_identity_registry_bridger_proxy_change_owner.sh` — run with `mainnet`
- `scripts/deployment/l2/script_l2_07_service_manager_proxy_change_owner.sh` — run with `<network>_mainnet`
- `scripts/deployment/l2/script_l2_08_identity_registry_bridger_proxy_change_owner.sh` — run with `<network>_mainnet`

## Behaviour

Each script reads the existing `globals_*.json` and calls `changeOwner(newOwner)` on the proxy, where `newOwner` is the chain's governance control point — the `timelockAddress` on Ethereum mainnet, the `bridgeMediatorAddress` on L2 (which is the bridge mediator, or the L1-aliased Timelock on Arbitrum). The scripts:

- pre-check that the signer derived from `derivationPath` is the current proxy owner, and bail out cleanly otherwise (avoids an `OwnerOnly` revert);
- are idempotent — a no-op if the proxy is already owned by the target;
- mirror the existing `deploy_21_22_service_staking_change_ownerships.sh` / tokenomics `script_02_buy_back_burner_change_owner.sh` patterns.

No globals or contract changes — these only add the four scripts.

### DIFF
--- a/scripts/deployment/l2/script_l2_07_service_manager_proxy_change_owner.sh
+++ b/scripts/deployment/l2/script_l2_07_service_manager_proxy_change_owner.sh
@@ -1,0 +1,100 @@
+#!/bin/bash
+
+# Transfers ServiceManagerProxy ownership from the deployer EOA to the chain's
+# governance control point, by calling changeOwner(bridgeMediator). On L2 this
+# is the bridge mediator; on Arbitrum it is the L1-aliased Timelock — both are
+# stored as bridgeMediatorAddress in the per-network globals.
+#
+# The script is idempotent: it pre-checks that the signer derived from
+# $derivationPath is the current proxy owner (otherwise changeOwner reverts with
+# OwnerOnly), and is a no-op if the proxy is already owned by the target.
+
+# Check if $1 is provided
+if [ -z "$1" ]; then
+  echo "Usage: $0 <network>"
+  echo "Example: $0 gnosis_mainnet"
+  exit 1
+fi
+
+red=$(tput setaf 1)
+green=$(tput setaf 2)
+reset=$(tput sgr0)
+
+# Get globals file
+globals="$(dirname "$0")/globals_$1.json"
+if [ ! -f $globals ]; then
+  echo "${red}!!! $globals is not found${reset}"
+  exit 0
+fi
+
+# Read variables using jq
+useLedger=$(jq -r '.useLedger' $globals)
+derivationPath=$(jq -r '.derivationPath' $globals)
+chainId=$(jq -r '.chainId' $globals)
+networkURL=$(jq -r '.networkURL' $globals)
+
+serviceManagerProxyAddress=$(jq -r '.serviceManagerProxyAddress' $globals)
+bridgeMediatorAddress=$(jq -r '.bridgeMediatorAddress' $globals)
+timelockAddress=$(jq -r '.timelockAddress' $globals)
+
+# New owner: bridge mediator on L2, Timelock on L1 mainnet.
+if [ "$bridgeMediatorAddress" != "null" ] && [ -n "$bridgeMediatorAddress" ]; then
+  newOwnerAddress="$bridgeMediatorAddress"
+elif [ "$timelockAddress" != "null" ] && [ -n "$timelockAddress" ]; then
+  newOwnerAddress="$timelockAddress"
+else
+  echo "${red}!!! Neither bridgeMediatorAddress nor timelockAddress is set in $globals${reset}"
+  exit 1
+fi
+
+if [ "$serviceManagerProxyAddress" == "null" ] || [ -z "$serviceManagerProxyAddress" ]; then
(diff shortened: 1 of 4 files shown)


DavidMinarsch:
## Code review

Nice, focused change — pure script additions, no contract or globals changes, addressing a real audit finding. I verified the contract side: `ServiceManagerProxy`/`IdentityRegistryBridgerProxy` delegatecall everything via `fallback()`, so `owner()` and `changeOwner(address)` resolve correctly against the implementation logic (`GenericManager`, `IdentityRegistryBridger`), both of which revert with `OwnerOnly` for a non-owner caller. The ABI calls are correct.

Two blocking issues, then some smaller points.

### 🔴 Scripts fail on bash 3.2

All four scripts use `${currentOwner,,}` / `${deployer,,}` (lowercase parameter expansion), which is **bash 4.0+ only**. The shebang is `#!/bin/bash`, and macOS ships bash **3.2.57** as `/bin/bash` by default — on which this is a runtime `bad substitution` error. The script would derive the deployer (possibly prompting the Ledger), then crash on the pre-flight `if` before `cast send` — broken, though no funds at risk. No other script in `scripts/` uses `,,`; this is new.

Lowercasing *is* needed for correctness — `newOwnerAddress` comes raw from the globals JSON and may not be EIP-55 checksummed, so it won't byte-match `cast`'s checksummed output. So don't just drop `,,`; replace it with something portable:

```bash
currentOwnerLc=$(echo "$currentOwner" | tr '[:upper:]' '[:lower:]')
```

### 🔴 Not actually idempotent — re-run errors with exit 1

The description claims "idempotent — a no-op if the proxy is already owned by the target." It isn't, because the two pre-flight checks are in the wrong order:

```bash
# Check 1: signer must be current owner
if [ "${currentOwner,,}" != "${deployer,,}" ]; then ... exit 1; fi
# Check 2: no-op if already owned by target
if [ "${currentOwner,,}" == "${newOwnerAddress,,}" ]; then ... exit 0; fi
```

After a successful transfer, `currentOwner` is the target (Timelock / bridge mediator), so **Check 1 fires first** and exits `1` with a red error plus a misleading hint ("Set derivationPath … to the path that controls `<timelock>`" — that address isn't controlled by any derivation path). Check 2 is **unreachable**: reaching it implies `currentOwner == deployer`, so it could only be true if `deployer == newOwnerAddress`, which never holds for an EOA vs. a governance contract.

Fix: swap the order — no-op check first, signer check second:

```bash
if [ "$currentOwnerLc" == "$newOwnerLc" ]; then
  echo "${green}... already owned by $newOwnerAddress. Nothing to do.${reset}"; exit 0
fi
if [ "$currentOwnerLc" != "$deployerLc" ]; then
  echo "${red}!!! Signer ... is not the current owner ...${reset}"; exit 1
fi
```

This also fixes batch runs: as written, an already-transferred network aborts a multi-network loop with exit 1.

### 🟡 L1 scripts prefer `bridgeMediatorAddress` despite being Timelock-only

The L1 scripts carry the same `bridgeMediatorAddress`-first / `timelockAddress`-fallback logic as L2, yet their header comment says "to the protocol Timelock on Ethereum mainnet." This works *only because* `globals_mainnet.json` happens to lack `bridgeMediatorAddress`. For an irreversible mainnet ownership transfer, the L1 scripts should select `timelockAddress` directly rather than rely on that absence.

### 🟢 Smaller points

- **L2 `timelockAddress` fallback is a dangerous default.** On L2, `timelockAddress` is the *non-aliased* L1 Timelock — an address nobody controls on that chain. If `bridgeMediatorAddress` were ever missing, the L2 script would silently transfer ownership to an unreachable address. Every L2 mainnet globals currently has `bridgeMediatorAddress`, so it's latent — but the L2 scripts should *require* it and error out rather than fall back.
- **A reverted `cast send` isn't detected.** `result=$($castCmd); echo "$result" | grep status` exits 0 even on `status 0 (failed)`. Consistent with existing scripts, but for an ownership transfer consider asserting `status 1`.
- **`cast call` failure unhandled.** An RPC error leaves `currentOwner` empty, falling into the "not owner ()" branch with a confusing message.
- **Duplication.** Four ~99-line files; L1 vs L2 differ only in comment text and the usage example. Matches the existing `script_l1_*`/`script_l2_*` convention, so acceptable — just noting a single parametrized script would be cleaner.

### Verdict

Request changes — the two 🔴 issues are blocking: the scripts won't run on macOS's default bash, and they aren't idempotent despite the description saying so. Both fixes are small and localized to the pre-flight block.

PR #309, opened 2026-07-03
Split failed proposal 24 into proposal 24 (de-whitelist + guard) and proposal 25 (un-nominate), under the EIP-7825 gas cap

kupermind:
## Why

The original single 41-action bundle (submitted as proposal 24) reverted on-chain: [`0x540c…20f`](https://etherscan.io/tx/0x540c2026c0a122465806240becdd8381489e19e61d74e74ec184a9660621e20f) failed with `status 0`, `gasUsed 16,530,883` against `gasLimit 16,777,216`.

That gas limit is exactly **2²⁴ = 16,777,216** — the **EIP-7825** (Fusaka) per-transaction gas cap. The bundle's `execute()` needs **~19.9M gas** (profiled on a mainnet fork), which no single transaction can supply. Every simulation (forge fork tests, Safe/Tenderly) passed because they run with a high gas allowance and **none enforce the per-tx cap** — only mainnet consensus does.

A GovernorOLAS `execute()` is atomic, so the bundle can't be executed in pieces and re-submitting fails identically. The fix is to split it into two independent, sub-cap proposals.

### Where the gas went (mainnet-fork profile)
| group | gas | note |
|---|---:|---|
| 4× OP-stack `sendMessage` (Optimism/Base/Celo/Mode) | ~9.57M | ~2.39M each — L1 gas-burn ∝ the 2M `minGasLimit` for a trivial L2 sstore |
| 32× `removeNominee` | ~9.55M | ~298k each (routes through the Dispenser) |
| Gnosis/Polygon/Arbitrum/mainnet de-whitelist + GuardCM | ~0.74M | |
| **total** | **~19.9M** | over the 16,777,216 cap |

## What

Two independent proposals, each measured (pure Timelock `execute`) well under the cap:

- **Proposal 24 — `scripts/proposals/proposal_24_dewhitelist_and_guard/`** (9 actions, **~10.4M gas**): de-whitelist the 8 same-address multisig adapters (mainnet + 7 L2 bridges; Polygon carries two adapters) + extend the GuardCM allowlist with 19 emergency-pause/drain triples. Carries the Arbitrum retryable value.
- **Proposal 25 — `scripts/proposals/proposal_25_unnominate/`** (32 actions, **~9.2M gas**): `VoteWeighting.removeNominee` for the 32 retired staking nominees. All direct L1 calls — no L2 propagation.

Bridge encodings, the GuardCM batch, and the Arbitrum retryable params are **byte-for-byte identical** to the original bundle, so L2 delivery is unchanged; only the L1 bundling split. Each proposal's description is self-contained and context-specific (24 = security hardening + guard; 25 = staking nominee cleanup).

**proposalIds**
- 24: `0xa670d15c929979d29c817dd94cf28995248de5b23b68dfa3403b624f173a125c`
- 25: `0xdaea1b75e72e86880322782b4a8a4a022992711bbeba32df08bc1d67feebbca6`

## Tests — all green (mainnet + L2 forks)

- `test/Proposal24ForkL1.t.sol` — full propose → vote → queue → execute through the live GovernorOLAS; asserts the mainnet de-whitelist + all 19 GuardCM triples; Arbitrum retryable executor-funded. **3 pass**
- `test/Proposal24ForkL2OpStack.t.sol` — Optimism/Base/Celo/Mode delivery. **4 pass**
- `test/Proposal24ForkL2Other.t.sol` — Gnosis (AMB) / Polygon (FxRoot, both adapters) / Arbitrum (aliased Timelock). **3 pass**
- `test/Proposal25ForkL1.t.sol` — full lifecycle; asserts all 32 nominees removed. **3 pass**

Pure-execute gas from the fork tests: proposal 24 `execute` = **10,380,834**, proposal 25 `execute` = **9,163,012** — both comfortably under 16,777,216.

## Timing note (proposal 25)

`removeNominee` reverts within the last week of the ongoing Tokenomics epoch (`block.timestamp >= getEpochEndTime(eCounter-1) + epochLen - 1 week`). The full propose→execute wall-clock is **~6.4 days** (votingDelay + votingPeriod ≈ 4.55 d **+ `governorDelay` ≈ 1.82 d**, not the Timelock `minDelay` of 0). Proposal 25 must be timed so execution lands in an epoch's first-week window.

## Cleanup

Removes the obsolete single-bundle proposal folder and its three fork tests (preserved in git history). The failed proposal can never be re-executed.

## Note on `MIN_GAS`

The 2M `minGasLimit` on the four OP-stack sends (a trivial L2 sstore) accounts for ~9.6M of the original total and is intentionally left unchanged here, to keep proposal 24's bridged payloads byte-identical to the already-simulated original bundle. It could be trimmed in a future revision for cheaper execution; the split alone resolves the cap violation.


77ph (changes requested):
**Audit review — requesting changes (one blocking item; the split itself is correct).**

The gas-cap split is sound and well done: the root cause is correctly diagnosed (the original 41-action bundle reverted at `gasUsed 16,530,883` against `gasLimit 16,777,216` = 2²⁴, the EIP-7825/Fusaka per-tx cap; the note that forge/Safe/Tenderly simulations all pass because none enforce the per-tx cap is the sharp catch), the 41 → 9 + 32 split is faithful with byte-for-byte-identical encodings, and the fork tests are green. **Proposal 24** (same-address adapter de-whitelists + GuardCM extension) is good.

**Blocking — proposal 25 (`removeNominee` × 32).** I checked the current weight of all 32 nominees on the live `VoteWeighting` (`0x9541…c5c1`):
- **31 of 32 have zero current weight** → clean removals.
- **1 — `n/a id2` (Base, chainId 8453, `0x51c5f4982B9b0B3c0482678f5847EA6228Cc8E54`) — still carries a substantial active vote (~64% of `getWeightsSum`).**

Removing a nominee that still has active votes leaves orphaned voting power in the sum accounting (governance vulnerabilities-list **item #8**). Please **exclude `n/a id2` from proposal 25**: execute the 31 zero-weight removals now, and handle `n/a id2` separately — e.g. have its voter(s) `revokeRemovedNomineeVotingPower` first so its removal is a clean zero-weight case. This is only about the proposal-25 nominee set; the split mechanics are unaffected. I'll approve once `n/a id2` is out of the batch.

77ph:
**Re-review of the post-review commits (`bdddb0ee9` — de-whitelist OLAS mech factories on 7 chains; `4ead3130a` — per-chain message batching) — verified first-hand.**

**Gas (the constraint that reverted the original 41-action bundle) — both proposals sit comfortably under the EIP-7825 per-tx cap (16,777,216):**

| Proposal | `execute()` gas | vs. cap | margin |
|---|---|---|---|
| **24** (11 actions) | **10,602,040** — asserted `< cap` in `Proposal24ForkL1` | ✓ | ~6.18M |
| **25** (32 `removeNominee`) | **9,163,050** — measured via the execute-as-Timelock path | ✓ | ~7.6M |

**Fork tests: 14/14 pass** (6 L1 + 8 L2) — full `propose → vote → queue → execute` on a mainnet fork for both proposals, and L2 combined-message relay+execute across all 7 chains. The batched per-chain messages (registry de-whitelist + factory de-whitelist in a single bridge send) execute for **~44k–160k gas** — far under the OP-stack `MIN_GAS = 2,000,000`. (`test_L2_gnosis` needs an archive-capable fork RPC — `https://rpc.gnosischain.com` — where it passes; the earlier failure was RPC state availability, not a code issue.)

**On-chain authorization re-verified on all 7 chains:** each Mech Marketplace `owner()` == the claimed governance mediator, and each OLAS factory is currently whitelisted (`mapMechFactories == 1`), so every `setMechFactoryStatuses([factory], [false])` is a correctly-authorized, meaningful de-whitelist. The cross-chain address coincidences (e.g. `MF_POLYGON` == the mainnet Treasury address, `SRL2_BASE` == the mainnet Timelock address) are deterministic-deployment artifacts — the contracts exist and behave correctly at those addresses on their own chains.

**Minor (test parity):** `Proposal25ForkL1` has no explicit `assertLt(executeGas, 16777216)`, unlike `Proposal24ForkL1`. The current execute is ~9.16M with large headroom, but the test does not guard against a future regression (e.g. more nominees). Suggest adding a symmetric gas-cap assertion to the proposal-25 fork test.

**Blocker unchanged — requesting changes stands.** Re-scanned the current weight of all 32 nominees on the live `VoteWeighting`: **31 are zero-weight; one — `n/a id2` (Base 8453, `0x51c5f4982B9b0B3c0482678f5847EA6228Cc8E54`) — still holds ~64% of `getWeightsSum`.** Removing an active-vote nominee leaves orphaned voting power (vulnerabilities-list item #8). Please still **exclude `n/a id2` from proposal 25** — execute the 31 zero-weight removals now, and handle `n/a id2` separately (e.g. `revokeRemovedNomineeVotingPower` on its voter(s) first, so its removal becomes a clean zero-weight case).

*Note (monitoring, not a blocker):* once proposal 24 executes, the OLAS-payment balance trackers on these 7 chains become retired branches — per the `autonolas-marketplace` vulnerabilities-list #17, deposits to a retired tracker are stuck (no payment path and no requester-withdrawal path), so a non-zero balance there should be read as stuck funds, not payment activity.

dagacha:
Blocking: `proposal_25_unnominate` still includes the active Base nominee `0x51c5f4982B9b0B3c0482678f5847EA6228Cc8E54` (`n/a id2`) in the 32-call `removeNominee` batch. This is the exact case called out in the previous review; removing it now would leave orphaned voting power in `VoteWeighting`. Please drop it from proposal 25 and resubmit only after the nominee is zero-weight, then update the proposal/test/docs counts to 31.\n\nAlso: `Proposal25ForkL1.t.sol` still doesn’t appear to have the explicit `executeGas < 16_777_216` guard that proposal 24’s fork test has. Not blocking, but worth adding once the batch is trimmed.

dagacha (approved):

77ph (approved):

PR #310, opened 2026-08-20
docs: add confirmed findings to the registries Vulnerabilities list (items 24-27)

kupermind:
Documentation only — **no contract source is changed**.

## What this records

`GnosisSafeSameAddressMultisig` and `StakingBase` validate a service multisig by comparing `multisig.codehash` against a configured `proxyHash`. That comparison proves the account holds Safe-proxy bytecode. It does **not** prove which singleton the proxy delegates to:

1. **The singleton is not in the code.** A Safe proxy stores its singleton in storage slot 0 and loads it at run time — the runtime bytecode carries an address *mask* (`PUSH20 0xff…ff`) and an `SLOAD(0)`, not the singleton address. Every proxy from a given factory therefore has byte-identical runtime code and an identical `codehash`, whatever singleton it points at.
2. **The factory does not constrain it.** `createProxyWithNonce(address _singleton, bytes initializer, uint256 saltNonce)` takes the singleton as a caller-supplied parameter on a permissionless factory, so "deployed by the canonical Safe factory" implies nothing about where a proxy delegates.

## Why it is not exploitable

Singleton identity is guaranteed **upstream, by construction**, not by the `proxyHash` check. `service.multisig` is only ever assigned from a whitelisted implementation's `create()`, and every whitelisted implementation builds the multisig itself against its own pinned constants:

| Implementation | Pins | Caller-supplied address? |
|---|---|---|
| `GnosisSafeMultisig` | `immutable gnosisSafe` + `immutable gnosisSafeProxyFactory` | No — fresh proxy |
| `SafeMultisigWithRecoveryModule` | `immutable safe` + `immutable safeProxyFactory` | No — fresh proxy |
| `PolySafeCreatorWithRecoveryModule` | pinned factory + `polySafeProxyBytecodeHash`, deterministic address | No — fresh proxy |
| `RecoveryModule.create` | reads the deploying service's own `mapServices` multisig, owners must equal its agent instances | No — re-verification |
| `GnosisSafeSameAddressMultisig` / `PolySafeSameAddressMultisig` | — | **Yes** — but de-whitelisted everywhere |

So a proxy with a non-canonical singleton can be created by anyone at any time, and there is no function that will accept it as a service multisig.

## Why record it anyway

The load-bearing control is the pinned `immutable` in each creating implementation, and that is not visible from either `codehash` call site. A future multisig implementation that adopts a **caller-supplied** address would restore the gap, and whoever writes it may reasonably assume `proxyHash` already covers implementation identity. The entry states the requirement explicitly: validate the singleton (implementation slot / `masterCopy()`) in addition to the `codehash` check.

## README

Marks `GnosisSafeSameAddressMultisig` and `PolySafeSameAddressMultisig` as **deprecated**. `mapMultisigs` returns `false` for both on Ethereum, Gnosis, Polygon, Base, Optimism, Arbitrum, Celo and Mode — verified on-chain — so `deploy()` reverts `UnauthorizedMultisig` if either is passed. The sources are deliberately retained: the contracts remain deployed on all eight chains, and the repo is the record of that deployed bytecode.

## Checks

Docs-only change, so the compile / test / lint / coverage jobs are unaffected by its content. No `.sol`, no ABI, no deployment artifact touched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,21 @@ Once all agent instances are registered, the service owner re-deploys the servic
 The implementation of such multisig is provided here:
 - [GnosisSafeSameAddressMultisig](./contracts/multisigs/GnosisSafeSameAddressMultisig.sol)
 
+> **Deprecated.** `GnosisSafeSameAddressMultisig` (and its Polygon counterpart
+> `PolySafeSameAddressMultisig`) are **no longer whitelisted on any supported chain** —
+> `mapMultisigs` returns `false` for them on Ethereum, Gnosis, Polygon, Base, Optimism, Arbitrum,
+> Celo and Mode, so `deploy()` reverts `UnauthorizedMultisig` if either is passed. The contracts
+> remain deployed and their sources are retained here as the record of that deployed bytecode, but
+> they must not be used for new service deployments. Use
+> [SafeMultisigWithRecoveryModule](./contracts/multisigs/SafeMultisigWithRecoveryModule.sol) or
+> [RecoveryModule](./contracts/multisigs/RecoveryModule.sol) instead.
+>
+> These are also the only multisig implementations that could ever set a service multisig to a
+> **caller-supplied existing** address; every other implementation creates a fresh Safe proxy against
+> its own pinned singleton. See item 24 in
+> [Vulnerabilities_list_registries.md](./docs/Vulnerabilities_list_registries.md) for why that
+> distinction matters.
+
 The updated version with the access recovery feature is provided in the Recovery Module contract itself here:
 - [RecoveryModule](./contracts/multisigs/RecoveryModule.sol)
 
--- a/docs/Vulnerabilities_list_registries.md
+++ b/docs/Vulnerabilities_list_registries.md
@@ -27,6 +27,7 @@
     - [21. calculateStakingLastReward rounding dust](#21-calculatestakinglastreward-rounding-dust)
     - [22. ServiceRegistry registerAgents / update / activateRegistration missing reentrancy guard](#22-serviceregistry-registeragents--update--activateregistration-missing-reentrancy-guard)
     - [23. deploy same-address multisig takeover (RecoveryModule)](#23-deploy-same-address-multisig-takeover-recoverymodule)
+    - [24. Safe proxy-hash validation checks proxy shape, not singleton identity](#24-safe-proxy-hash-validation-checks-proxy-shape-not-singleton-identity)
 
 ## Involved contracts and level of the bugs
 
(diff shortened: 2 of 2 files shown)


DavidMinarsch (commented):
Verified the claims against live chain state and the contracts rather than reading the prose. **Everything checks out, including the part that is easiest to assert and hardest to confirm** — that the `proxyHash` gap has no reachable path.

## Verified on-chain

**The de-whitelisting is exact.** `mapMultisigs` returns `false` for all nine deployments across the eight chains:

| | Ethereum | Gnosis | Polygon | Base | Optimism | Arbitrum | Celo | Mode |
|---|---|---|---|---|---|---|---|---|
| `GnosisSafeSameAddressMultisig` | false | false | false | false | false | false | false | false |
| `PolySafeSameAddressMultisig` | — | — | false | — | — | — | — | — |

And the replacements the README points at are live: `SafeMultisigWithRecoveryModule`, `RecoveryModule` and `GnosisSafeMultisig` all return `true`. So the deprecation note is actionable, not aspirational.

**#24's mechanism claim is exact, byte for byte.** A live Safe proxy's runtime code begins:

```
6080604052 73 ffffffffffffffffffffffffffffffffffffffff 600054 16 7f…
           └ PUSH20 mask ────────────────────────────┘ └SLOAD(0)┘└AND┘
```

The singleton sits in slot 0 (`0xd9db270c…` on the one I checked) and its bytes are **absent from the runtime code**. I also found two distinct proxy `codehash` values in current use (170- and 171-byte variants from different factory versions), both carrying the singleton in storage — which is the same point from the other direction: `codehash` tracks the proxy variant, never the delegate.

**The "not exploitable" argument is complete against live state — this is the claim I would most want checked, so I checked it.** I enumerated every implementation in `configuration.json` and queried `mapMultisigs` live. The whitelisted set is exactly:

`GnosisSafeMultisig` · `RecoveryModule` · `SafeMultisigWithRecoveryModule` · `PolySafeCreatorWithRecoveryModule` (Polygon)

— precisely the four "No — fresh proxy / re-verification" rows of the entry's table, with nothing missing. So there is no whitelisted implementation the argument overlooks, and the conclusion holds against deployed state, not just against the source.

**#27's asymmetry is real.** `unbondWithSignature` reads and writes `mapOperatorUnbondNonces` (`ServiceManager.sol:588`, `:597`); the direct `unbond()` path (`:532-564`) contains zero references to it.

Anchors for all four new entries resolve, and the severities (2 Informative, 2 Low) sit comfortably inside the document's own distribution (14 Low / 10 Informative / 2 High).

## The guidance in #24 is the valuable part, and it is in the wrong place

The entry's own framing is that the risk is prospective: *"a future multisig implementation that adopts a caller-supplied address would restore the gap, and whoever writes it may reasonably assume the `proxyHash` comparison already covers implementation identity."*

That person will be writing a new multisig implementation and getting it added to `mapMultisigs`. There is no particular reason they will read `Vulnerabilities_list_registries.md` first — the whole premise is that they hold a *wrong assumption they do not know to check*. A vulnerabilities-list entry only reaches someone already looking for known issues.

The requirement would bind where the mistake gets made: a comment at the `proxyHash` comparison in `GnosisSafeSameAddressMultisig` and `StakingBase` saying what it does **not** prove, and/or a line in whatever review step gates additions to `mapMultisigs`. Cheap, and it puts the constraint in front of the person who would otherwise break it.

## The deprecation is now load-bearing for two entries

Re-whitelisting either same-address implementation would simultaneously reopen **#24** (the only implementations that accept a caller-supplied address) and **#23** (`deploy` same-address multisig takeover). Both entries currently rest on the same `mapMultisigs = false` state, which is a governance-mutable flag rather than a code property.

Worth saying so explicitly in #24 — a future proposal to re-whitelist should be treated as a security change requiring the singleton validation to land first, not as a configuration tweak.

## Minor

- **#26 and #27 are two symptoms of one shape**: the direct-unbond path does not perform cleanup that the alternative path does — the bridger mapping in one case, the nonce in the other. Neither entry references the other, though a lifecycle fix would naturally address both. A cross-reference would stop the second being forgotten once the first is fixed.
- **`Informational` vs `Informative`.** The document uses `Informative` ten times and `Informational` once (line 285). Pre-existing, not introduced here — but this PR is touching the file and the severity strings are what a reader greps.

PR #321, opened 2026-08-31
docs: fix three dead vulnerabilities-list links, add the bounty link, correct the licence

DavidMinarsch:
Documentation hygiene. No contract changes.

## Three dead links in the public vulnerabilities list

- **`contracts/ServiceManagerToken.sol` (×2)** — renamed to `ServiceManager.sol`. The stale `#L120` anchor on the second is **dropped rather than guessed** at a new line: the zero-value check has moved within the file, and a wrong anchor is worse than none.
- **`docs/internal_audit_4.pdf`** — wrong in both parts. Audits live in `audits/`, not `docs/`, and no file of that name exists anywhere in the repo. Now points at [`audits/internal4/README.md`](audits/internal4/README.md), which is the audit it meant.

## Also

- **Bounty link** added — it was reachable from the governance README only, though the programme covers the contracts here too.
- **Licence metadata:** `package.json` said `"license": "ISC"` while the repo ships an **MIT** `LICENSE` and MIT SPDX headers.

Verified: 61 relative links across README and `docs/`, **0 broken**.

### DIFF
--- a/docs/Vulnerabilities_list_registries.md
+++ b/docs/Vulnerabilities_list_registries.md
@@ -114,8 +114,8 @@ would not be affected by any possible slashing condition if the total operator b
 equal to zero.
 
 This vulnerability is addressed for the ServiceRegistry contract and ServiceRegistryL2 by
-adding the zero-value check on the service manager level. Specifically, [serviceManager](../contracts/ServiceManagerToken.sol)
-contract handles the [check](../contracts/ServiceManagerToken.sol#L120) before calling the original serviceRegistry's update() method.
+adding the zero-value check on the service manager level. Specifically, [serviceManager](../contracts/ServiceManager.sol)
+contract handles the [check](../contracts/ServiceManager.sol) before calling the original serviceRegistry's update() method.
 See ../test/ServiceManagerToken.js#L326-L333C25
 for a test proving that the issue is resolved.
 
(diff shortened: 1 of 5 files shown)


docs/Vulnerabilities_list_registries.md
```diff
@@ -114,8 +114,8 @@ would not be affected by any possible slashing condition if the total operator b
 equal to zero.
 
 This vulnerability is addressed for the ServiceRegistry contract and ServiceRegistryL2 by
-adding the zero-value check on the service manager level. Specifically, [serviceManager](../contracts/ServiceManagerToken.sol)
-contract handles the [check](../contracts/ServiceManagerToken.sol#L120) before calling the original serviceRegistry's update() method.
+adding the zero-value check on the service manager level. Specifically, [serviceManager](../contracts/ServiceManager.sol)
```
DavidMinarsch, line 117: Not sure that's correct - please check

DavidMinarsch: Checked — the change is valid, and here is the evidence rather than my word for it.

## It is a rename, not a different contract

```
$ git show cdc37b5 --name-status -M
R100    contracts/ServiceManagerToken.sol    contracts/ServiceManager.sol
```

**`R100` is a 100% similarity rename** — same file, new name, zero content change in that commit. `git log --follow contracts/ServiceManager.sol` traces straight back through it, and the two never coexisted at any commit.

## The deployed contract is called `ServiceManager`

`docs/configuration.json` — which is the record of what is actually deployed — names it `ServiceManager` on **all eight chains** (mainnet, Polygon, Gnosis, Arbitrum, Optimism, Base, Celo, Mode). Nothing called `ServiceManagerToken` is deployed anywhere. So the link now matches both the current source *and* the deployed name.

## The check is really in there

Not just the file — the specific check the entry describes:

```solidity
// ServiceManager.sol:298-302, inside update()
if (agentParams[i].slots > 0 && agentParams[i].bond == 0) {
    revert ZeroValue();
}
```

with the equivalent on the token path immediately below. I have pinned the second link to `#L298-L302` rather than leaving it at the file, since the original `#L120` anchor was stale.

## One thing I deliberately did not change

The prose still points at **`../test/ServiceManagerToken.js`** — and that is correct, because **the test file was never renamed**. It still exists under the old name. So the entry now reads `ServiceManager.sol` for the contract and `ServiceManagerToken.js` for the test, which looks inconsistent but is accurate. Renaming the test file is a separate call and not one for a docs PR; say the word if you want it done.

PR #322, opened 2026-09-01
docs: add registries vulnerability 35

DavidMinarsch:
**35 — Staking does not verify the instance it stakes into** (Low). `verifyInstance` does not appear in `StakingBase` at all, so an instance the factory no longer verifies can still accept stakes and pay rewards.

Whether that yields anything depends on the instance's activity checker — a checker keyed to a Safe-level marker rather than an instance-level one would let a service that never registered with the instance satisfy liveness. The entry states the live exposure: of the funded-but-unnominated contracts only two fail `verifyInstance()`, and neither uses a checker of that kind. The gap is worth recording because instance verification, not checker choice, ought to be the control.

### DIFF
--- a/docs/Vulnerabilities_list_registries.md
+++ b/docs/Vulnerabilities_list_registries.md
@@ -39,6 +39,7 @@
     - [33. Staking snapshots are suppressed while the reward reserve is empty](#33-staking-snapshots-are-suppressed-while-the-reward-reserve-is-empty)
     - [34. `RecoveryModule.create` accepts a Safe that has no recovery module enabled](#34-recoverymodulecreate-accepts-a-safe-that-has-no-recovery-module-enabled)
 
+    - [35. Staking does not verify the instance it stakes into](#35-staking-does-not-verify-the-instance-it-stakes-into)
 ## Involved contracts and level of the bugs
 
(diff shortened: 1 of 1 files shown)


DavidMinarsch:
Withdrawing this entry — the premise is wrong, and the source says so.

`StakingFactory.verifyInstanceAndGetEmissionsAmount`, which is the function the incentive path actually uses, states the design decision in place:

```
// Check if the proxy instance is enabled, since this proves that
// the proxy instance has been created by this factory and verified at the creation time
// An already verified proxy instance should not be re-verified
// DAO governance might have changed verification rules in the mean-time which would render the instance unusable
```

Verification is deliberately creation-time only, and the live control is `mapInstanceParams[instance].isEnabled`, not `verifyInstance`. So `stake()` not calling `verifyInstance` is not a gap — calling it would be the defect, because it would render unusable every instance created under verifier rules that have since changed.

The two instances I cited as evidence confirm the design rather than a problem: both return `verifyInstance() == false` with `isEnabled == true`, and `verifyInstanceAndGetEmissionsAmount` returns a normal 1140 OLAS for each. That is exactly the situation the comment above anticipates.

PR #324, opened 2026-09-06
feat(l2): add Robinhood Chain (4663) deployment globals

DavidMinarsch:
Adds `scripts/deployment/l2/globals_robinhood_mainnet.json` for the Robinhood Chain rollout. Robinhood Chain is an Arbitrum Orbit rollup settling to Ethereum with ETH gas, so it takes the Arbitrum path with no contract changes anywhere.

Every value was verified against the live chains on 2026-09-06, and the file covers **all 34 globals keys** read by the in-scope L2 shell scripts (checked mechanically — no missing keys).

### Values worth a second look on review

- **`bridgeMediatorAddress` is byte-identical to Arbitrum's** (`0x4d30F68F…A70F`). That is correct, not a copy-paste error: it is the L2 alias of the same L1 Timelock, and address aliasing is chain-independent. Re-derive it as `Timelock + 0x1111…1111 mod 2^160` if in doubt.
- **`olasAddress` has no code yet.** `0x0929…2286` is the deterministic standard-gateway address for bridged OLAS; the first deposit through the L1 gateway router deploys it. Both the L1 and L2 routers return this value from `calculateL2TokenAddress`.
- **`multisigProxyHash130`** carries over from Arbitrum verbatim, because Safe 1.3.0's singleton, proxy factory and MultiSendCallOnly are all at their canonical addresses on 4663 (confirmed by code size).
- **The four `StakingVerifier` limits** match every other L2 exactly; only Ethereum differs, and only in `numServicesLimit`.

### Deliberate omissions

- `stakingNativeTokenAddress` is left empty and **`deploy_17` should not be run**. `StakingNativeToken` is deployed on Mode alone and is being disabled there, so there is nothing to mirror.
- No PolySafe keys — Polygon-only.
- No `hashCheckpointAddress` — absent on Arbitrum and Celo.

### Verification caveat

**Etherscan V2 does not cover chain 4663.** Each script tries Etherscan first and Blockscout second; the Etherscan leg will fail and Blockscout is what lands. The scripts neither `set -e` nor check that return, so this is noise rather than a failure — but expect it in the logs. `blockscoutURL` is set to `https://robinhoodchain.blockscout.com`.

Because of this the **shell/Forge route should be used end-to-end** rather than the Hardhat `.js` route, which would additionally need a `robinhood` network and an `etherscan.customChains` entry in `hardhat.config.js`. This also matches the existing convention from `OLAS1-6`.

No contract changes. Config only.


scripts/deployment/l2/globals_robinhood_mainnet.json
```diff
@@ -0,0 +1,44 @@
...
+  "gnosisSafeAddress": "0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552",
+  "gnosisSafeProxyFactoryAddress": "0xa6B71E26C5e0845f74c812102Ca7114b6a896AB2",
+  "multiSendCallOnlyAddress": "0x40A2aCCbd92BCA938b02010E17A5b8929b49130D",
+  "identityRegistryAddress": "0x8004A169FB4a3325136EB29fA0ceB6D2e539a432",
+  "reputationRegistryAddress": "0x8004BAa17C55a88189AE136b182e5fdA19dE9b63",
+  "validationRegistryAddress": "",
+  "baseURI": "https://gateway.autonolas.tech/ipfs/",
+  "serviceRegistryName": "Service Registry L2",
+  "serviceRegistrySymbol": "AUTONOLAS-SERVICE-L2-V1",
+  "timelockAddress": "0x3C1fF68f5aa342D296d4DEe4Bb1cACCA912D95fE",
```
DavidMinarsch, line 19: Validated against autonolas-governance repo

scripts/deployment/l2/globals_robinhood_mainnet.json
```diff
@@ -0,0 +1,44 @@
...
+  "useLedger": true,
+  "derivationPath": "m/44'/60'/2'/0/0",
+  "providerName": "robinhood",
+  "chainId": "4663",
+  "networkURL": "https://rpc.mainnet.chain.robinhood.com",
+  "blockscoutURL": "https://robinhoodchain.blockscout.com",
+  "gasPriceInGwei": "1",
+  "gnosisSafeAddress": "0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552",
+  "gnosisSafeProxyFactoryAddress": "0xa6B71E26C5e0845f74c812102Ca7114b6a896AB2",
+  "multiSendCallOnlyAddress": "0x40A2aCCbd92BCA938b02010E17A5b8929b49130D",
```
DavidMinarsch, line 12: Validated against https://docs.safe.global/advanced/smart-account-supported-networks?search=robinhood&expand=4663

scripts/deployment/l2/globals_robinhood_mainnet.json
```diff
@@ -0,0 +1,44 @@
...
+  "identityRegistryAddress": "0x8004A169FB4a3325136EB29fA0ceB6D2e539a432",
+  "reputationRegistryAddress": "0x8004BAa17C55a88189AE136b182e5fdA19dE9b63",
```
DavidMinarsch, line 14: Validated against https://8004scan.io/networks

scripts/deployment/l2/globals_robinhood_mainnet.json
```diff
@@ -0,0 +1,44 @@
...
+  "validationRegistryAddress": "",
+  "baseURI": "https://gateway.autonolas.tech/ipfs/",
+  "serviceRegistryName": "Service Registry L2",
+  "serviceRegistrySymbol": "AUTONOLAS-SERVICE-L2-V1",
+  "timelockAddress": "0x3C1fF68f5aa342D296d4DEe4Bb1cACCA912D95fE",
+  "bridgeMediatorAddress": "0x4d30F68F5AA342d296d4deE4bB1Cacca912dA70F",
```
DavidMinarsch, line 20: L2 alias of L1 timelock

scripts/deployment/l2/globals_robinhood_mainnet.json
```diff
@@ -0,0 +1,44 @@
...
+  "olasAddress": "0x092963938DEBd8013a2E545B3549f8a5EC0d2286",
+  "multisigProxyHash130": "0xb89c1b3bdf2cf8827818646bce9a8f6e372885f8c55e5c07acbd307cb133b000",
```
DavidMinarsch, line 22: As on other chains

scripts/deployment/l2/globals_robinhood_mainnet.json
```diff
@@ -0,0 +1,44 @@
...
+  "agent8004BaseURI": "https://marketplace.olas.network/erc8004/robinhood-chain/ai-agents/",
```
DavidMinarsch:
```suggestion
  "agent8004BaseURI": "https://marketplace.olas.network/erc8004/robinhood/ai-agents/",
```

scripts/deployment/l2/globals_robinhood_mainnet.json
```diff
@@ -0,0 +1,44 @@
...
+  "minStakingDepositLimit": "10000000000000000000000",
+  "timeForEmissionsLimit": "2592000",
+  "numServicesLimit": "100",
+  "apyLimit": "3000000000000000000",
```
DavidMinarsch, line 27: Same as other l2s - check autonolas-registries

DavidMinarsch:
Thanks — all the validations line up with what I read independently, and the `agent8004BaseURI` suggestion is applied (`86d36fe`): the slug is now `robinhood`, matching the chain key rather than the full name.

For the record on the two you flagged for provenance:

- **`bridgeMediatorAddress`** — re-derived from scratch as `Timelock + 0x1111…1111 mod 2^160` and it reproduces exactly, and it matches `bridgeMediatorAddress` in this repo's own `globals_arbitrum_mainnet.json`. Independently anchored too: `Dispenser.owner()` and `Depository.owner()` on L1 are both the same Timelock the alias derives from.
- **The four `StakingVerifier` limits** — identical across Arbitrum, Base, Celo, Gnosis, Mode, Optimism and Polygon. Only Ethereum differs, and only in `numServicesLimit` (500 vs 100).

Every address in the file was re-verified on-chain on 2026-09-06, and the file covers all 34 globals keys the in-scope L2 shell scripts read (checked by extracting the `jq -r` keys from the scripts rather than by eye).

scripts/deployment/l2/globals_robinhood_mainnet.json
```diff
@@ -0,0 +1,44 @@
...
+  "reputationRegistryAddress": "0x8004BAa17C55a88189AE136b182e5fdA19dE9b63",
+  "validationRegistryAddress": "",
+  "baseURI": "https://gateway.autonolas.tech/ipfs/",
+  "serviceRegistryName": "Service Registry L2",
+  "serviceRegistrySymbol": "AUTONOLAS-SERVICE-L2-V1",
+  "timelockAddress": "0x3C1fF68f5aa342D296d4DEe4Bb1cACCA912D95fE",
+  "bridgeMediatorAddress": "0x4d30F68F5AA342d296d4deE4bB1Cacca912dA70F",
+  "olasAddress": "0x092963938DEBd8013a2E545B3549f8a5EC0d2286",
+  "multisigProxyHash130": "0xb89c1b3bdf2cf8827818646bce9a8f6e372885f8c55e5c07acbd307cb133b000",
+  "agent8004BaseURI": "https://marketplace.olas.network/erc8004/robinhood/ai-agents/",
```
OjusWiZard, line 23: **`agent8004BaseURI` here contradicts the plan document, and this value is baked in at deploy time.**

```
this PR                    …/erc8004/robinhood/ai-agents/
plan rev 11, line 311      …/erc8004/robinhood-chain/ai-agents/
```

Two open PRs currently disagree on a value that ends up on-chain.

The slug is **not** simply `providerName` — it diverges on three of the seven existing chains: `arbitrum`→`arbitrum-one`, `mode`→`mode-mainnet`, `optimism`→`op-mainnet`. So the bare provider name is not a safe default; the slug tracks the marketplace's own routes, and the plan's `robinhood-chain` fits that pattern better than `robinhood` does.

Worth settling in this PR rather than later: `deploy_28_identity_registry_bridger_proxy.sh:33` passes it straight into `initialize(string)`. `changeBaseURI` exists but is `onlyOwner`, and plan §3.7 hands `IdentityRegistryBridgerProxy` to the bridge mediator as the last act of wave 1 — so this is **one EOA transaction to fix before the handover, and a governance proposal plus a cross-chain retryable after**. It also fails silently: a wrong slug just 404s agent metadata, with nothing in the acceptance checklist that would notice.

Please confirm against the marketplace's actual route and make the two PRs agree.

scripts/deployment/l2/globals_robinhood_mainnet.json
```diff
@@ -0,0 +1,44 @@
...
+  "multiSendCallOnlyAddress": "0x40A2aCCbd92BCA938b02010E17A5b8929b49130D",
+  "identityRegistryAddress": "0x8004A169FB4a3325136EB29fA0ceB6D2e539a432",
+  "reputationRegistryAddress": "0x8004BAa17C55a88189AE136b182e5fdA19dE9b63",
+  "validationRegistryAddress": "",
+  "baseURI": "https://gateway.autonolas.tech/ipfs/",
+  "serviceRegistryName": "Service Registry L2",
+  "serviceRegistrySymbol": "AUTONOLAS-SERVICE-L2-V1",
+  "timelockAddress": "0x3C1fF68f5aa342D296d4DEe4Bb1cACCA912D95fE",
+  "bridgeMediatorAddress": "0x4d30F68F5AA342d296d4deE4bB1Cacca912dA70F",
+  "olasAddress": "0x092963938DEBd8013a2E545B3549f8a5EC0d2286",
```
OjusWiZard, line 21: **This address fails its EIP-55 checksum.**

```
here       0x092963938DEBd8013a2E545B3549f8a5EC0d2286
correct    0x092963938deBD8013a2e545b3549f8A5ec0D2286
```

The address itself is right — case-insensitively it matches what both the L1 and L2 routers return from `calculateL2TokenAddress` — but the casing is invalid. It is the **only one of this file's ten addresses** that fails, so it reads as a transcription artifact rather than a convention.

It matters because the key has consumers on both routes. `deploy_18_service_staking_verifier.sh` reads it through `jq` into `cast`, which does not validate — fine. But `deploy_18_service_staking_verifier.js:66` and `verify_18_service_staking_verifier.js:5` hand it to ethers, whose `getAddress()` rejects a **mixed-case** address that fails EIP-55 (an all-lowercase one would pass). The PR recommends the shell route end-to-end, but `verify_18_*.js` is a `.js` helper reading this same key.

The same mis-cased string has propagated to all three artifacts — this PR, `autonolas-tokenomics#357`, and the plan doc — so it is worth fixing in one pass.

OjusWiZard (commented):
Reviewed by checksum-validating the file, diffing it field-by-field against `globals_arbitrum_mainnet.json`, re-deriving the derived values, and testing the coverage claim. Two inline comments below.

**The coverage claim checks out exactly.** I extracted the keys mechanically: **36** distinct keys are read across all `l2/*.sh`, of which exactly two — `polySafeProxyBytecodeHash` and `polySafeProxyFactoryAddress` — belong to the PolySafe scripts this PR deliberately excludes. 36 − 2 = **34**. No in-scope key is missing, and the key set is *identical* to Arbitrum's — nothing extra, nothing dropped.

Also independently confirmed: `bridgeMediatorAddress` re-derives exactly as `Timelock + 0x1111…1111 mod 2^160`; `multisigProxyHash130` matches Arbitrum verbatim; the four `StakingVerifier` limits match Arbitrum; the ERC-8004 addresses are byte-identical to Arbitrum's, so the cross-chain vanity claim holds; `validationRegistryAddress: ""` is consistent with all seven existing chains; `gasPriceInGwei: "1"` is sensible against the ~0.47 gwei I measured on 4663; and the remaining nine addresses are checksum-valid and present on-chain at the stated byte counts.

Both findings below are one-character-class fixes and neither changes the approach.

scripts/deployment/l2/globals_robinhood_mainnet.json
OjusWiZard: Correction to my own comment: I have since seen your (now-outdated) `suggestion` on this line moving it **to** `robinhood`, so this was a deliberate choice, not a slip. I had the direction backwards.

The substance stands but inverts: the two artifacts still disagree and the value still goes on-chain via `initialize(string)`, so it is worth reconciling — but it is the **plan doc (rev 11, line 311) that is stale**, not this file. Either update the plan to `robinhood`, or say why the slug departs from the `arbitrum-one` / `mode-mainnet` / `op-mainnet` pattern so the next reader does not re-raise it.

The EIP-55 point on line 21 is unaffected.

scripts/deployment/l2/globals_robinhood_mainnet.json
dagacha, line 20: Verified: this is exactly the L1 timelock (`0x3C1f…95fE`) plus the Arbitrum address-alias offset `0x1111…1111 mod 2^160`. I re-derived it and it matches byte-for-byte. Correct, not a copy-paste error. ✅

scripts/deployment/l2/globals_robinhood_mainnet.json
dagacha, line 21: **[Must fix] Invalid EIP-55 checksum.** I computed the canonical checksum with Keccak-256 (verified against the EIP-55 spec vectors):

```
given    0x092963938DEBd8013a2E545B3549f8a5EC0d2286   <- invalid
correct  0x092963938deBD8013a2e545b3549f8A5ec0D2286
```

This is the only one of the file's 8 populated addresses that fails, so it reads as a transcription artifact. The `.sh` route tolerates it (jq → `cast abi-encode` doesn't validate), but **`deploy_18_service_staking_verifier.js` and `verify_18_service_staking_verifier.js` hand it to ethers, whose `getAddress()` rejects a mixed-case address that fails EIP-55** — a latent landmine even if the shell route is used end-to-end. Fix to the checksummed form above.

scripts/deployment/l2/globals_robinhood_mainnet.json
dagacha, line 23: **[Needs confirmation] Slug `robinhood` vs the plan's `robinhood-chain`.**

The slug does **not** simply equal `providerName` — it diverges on 3 of the 7 existing chains:

```
arbitrum → arbitrum-one   mode → mode-mainnet   optimism → op-mainnet
base/celo/gnosis/polygon → same as providerName
```

So the bare provider name is not a reliable default, and the plan doc (rev 11, line 311) says `robinhood-chain`. This value is baked on-chain via `deploy_28_identity_registry_bridger_proxy.sh:33 → initialize(string)`, and is `onlyOwner`-changeable only *after* the bridger proxy is handed to the bridge mediator — expensive to fix later, and a wrong slug fails silently (agent metadata just 404s). Please confirm against the live marketplace route and align the two open PRs so they don't disagree on a value that ends up on-chain.

scripts/deployment/l2/globals_robinhood_mainnet.json
```diff
@@ -0,0 +1,44 @@
...
+  "serviceManagerProxyAddress": "",
+  "gnosisSafeMultisigImplementationAddress": "",
+  "gnosisSafeSameAddressMultisigImplementationAddress": "",
+  "recoveryModuleAddress": "",
+  "safeMultisigWithRecoveryModuleAddress": "",
+  "complementaryServiceMetadataAddress": "",
+  "identityRegistryBridgerAddress": "",
+  "identityRegistryBridgerProxyAddress": "",
+  "stakingTokenAddress": "",
+  "stakingNativeTokenAddress": "",
```
dagacha, line 41: **[Info / guard] `deploy_17` must not be run.** Note that `deploy_17_service_staking_native_token.sh` does **not** read this key from globals — it deploys `StakingNativeToken` and writes the address back. So "don't run deploy_17" is enforced only by operator discipline, not by the config. Consider a short comment/guard here so a future operator can't silently deploy a staking-native contract that's being disabled on Mode.

dagacha (commented):
Config-only change — no contract changes. I independently re-derived the numeric claims rather than taking them on trust; overall the values are sound, with two items to settle before merge.

_This review was performed with **Pi** (pi-coding-agent) using the **Qwen3.8 Max** model (Aliyun Bailian token-plan provider) at **high** reasoning effort._

**Independently verified ✅**
- `bridgeMediatorAddress` is exactly the L1 timelock + Arbitrum address-alias offset — matches byte-for-byte.
- All populated addresses pass EIP-55 **except** `olasAddress` (see inline).
- Key coverage matches the keys read by the in-scope `deploy_*.sh` scripts; `polySafe*` (Polygon-only) and `hashCheckpointAddress` are correctly omitted, consistent with the Arbitrum/Celo baseline.

**Blocking before merge**
1. Fix the `olasAddress` EIP-55 checksum (line 21) — it's the only invalid one and ethers-based helpers (`deploy_18.js` / `verify_18.js`) will reject it.
2. Confirm the `agent8004BaseURI` slug (line 23) against the live marketplace route — it diverges from `providerName` on 3 of 7 chains, and this value is baked on-chain and expensive to change later.

**Recommended (non-blocking)**
- Add a guard/comment on `stakingNativeTokenAddress` since `deploy_17` must not be run (line 41).

Inline detail below.

scripts/deployment/l2/globals_robinhood_mainnet.json
```diff
@@ -0,0 +1,44 @@
...
+  "derivationPath": "m/44'/60'/2'/0/0",
+  "providerName": "robinhood",
+  "chainId": "4663",
+  "networkURL": "https://rpc.mainnet.chain.robinhood.com",
+  "blockscoutURL": "https://robinhoodchain.blockscout.com",
+  "gasPriceInGwei": "1",
+  "gnosisSafeAddress": "0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552",
+  "gnosisSafeProxyFactoryAddress": "0xa6B71E26C5e0845f74c812102Ca7114b6a896AB2",
+  "multiSendCallOnlyAddress": "0x40A2aCCbd92BCA938b02010E17A5b8929b49130D",
+  "identityRegistryAddress": "0x8004A169FB4a3325136EB29fA0ceB6D2e539a432",
```
dagacha, line 13: Verified these dependencies with read-only RPC calls on chain 4663. The Safe singleton, proxy factory and MultiSendCallOnly runtime bytecode each match the corresponding Arbitrum deployment byte-for-byte. Both ERC-8004 proxy addresses have deployed implementations: the identity implementation has 14,474 bytes of code and the reputation implementation has 10,491 bytes. No issue found in these dependency-address checks.

scripts/deployment/l2/globals_robinhood_mainnet.json
dagacha, line 19: Independently re-derived this address as `(timelockAddress + 0x1111000000000000000000000000000000001111) mod 2^160`; it matches exactly. Robinhood documents L1-to-L2 address aliasing, so reusing the Arbitrum alias for the same L1 Timelock is consistent with the intended governance path. [Robinhood address-aliasing documentation](https://docs.robinhood.com/chain/differences-from-ethereum/).

scripts/deployment/l2/globals_robinhood_mainnet.json
dagacha, line 21: Verified both derived values against the live Robinhood chain: `calculateL2TokenAddress(0x0001A500A6B18995B03f44bb040A5fFc28E45CB0)` on the documented L2 Gateway Router (`0x1E324B9316138CA9a73F960213621AD1aaf01B89`) returns the configured OLAS address bytes, and `keccak256(proxyRuntimeCode())` on the configured Safe factory equals `multisigProxyHash130`. The OLAS address currently returns empty code, consistent with the PR’s disclosed pre-deposit state. [Official gateway address](https://docs.robinhood.com/chain/protocol-contracts/).

dagacha (commented):
No actionable findings in the config-only change at `86d36fe56f37b7a9cb5ef83bec8e264a3773f04e` for the stated shell/Forge deployment scope.

Review performed with Codex using **gpt-6-astra** at **high** reasoning effort (confirmed from this review session’s metadata).

Validated the JSON and mechanically checked all 34 globals keys consumed by the in-scope L2 shell scripts: none are missing. The four staking limits match the existing L2 mainnet configurations. The chain ID, RPC URL and explorer agree with [Robinhood’s network documentation](https://docs.robinhood.com/chain/connecting/), and the RPC itself returns chain ID 4663.

The address and bytecode checks are recorded inline. Validation consisted of static inspection and read-only live RPC calls; no deployment or end-to-end deployment test was performed. The configured OLAS address currently has no code, as disclosed in the PR. This review follows the PR’s stated Forge-only scope and does not establish Hardhat compatibility or validate the marketplace metadata endpoint.

bennyjo (dismissed):
Approving. Independently verified every fixed value on-chain (Safe 1.3.0 set, proxy hash recomputed from the deployed proxy code, ERC-8004 registries, OLAS via both gateway routers, Timelock alias) and confirmed all 34 shell-script keys are present. Please fix the `olasAddress` checksum already raised before merging. Two notes on the description: OLAS now has code at the listed address, and Robinhood's Blockscout API returns a Cloudflare 403 to non-browser clients, so the Blockscout verify leg fails as well as the Etherscan one; plan a manual verification path for the run.

DavidMinarsch:
Both fixed in `d91e33f`. CI green.

**Checksum** — `olasAddress` corrected to `0x092963938deBD8013a2e545b3549f8A5ec0D2286`. Fixed in all four artifacts in one pass, as suggested.

**The slug goes back to `robinhood-chain`, and my earlier suggestion-application was wrong.** @OjusWiZard and @dagacha were both right to push on this, and it turned out to be settleable rather than a judgement call.

It isn't a free choice at all. `autonolas-frontend-mono` computes it in `apps/marketplace/common-util/Login/config.tsx`:

```ts
networkName: kebabCase(name),   // name comes from the wagmi/viem chain object
```

and `erc8004Helpers.ts` resolves the route's `[network]` segment by matching that value:

```ts
const chain = EVM_SUPPORTED_CHAINS.find((c) => c.networkName.toLowerCase() === network.toLowerCase());
```

So the slug is `kebabCase(viemChain.name)`. That rule reproduces every existing slug exactly — `'Arbitrum One'`→`arbitrum-one`, `'OP Mainnet'`→`op-mainnet`, `'Mode Mainnet'`→`mode-mainnet`, and base/celo/gnosis/polygon where the name already equals the key. viem's chain 4663 is `name: 'Robinhood Chain'`, so the slug is **`robinhood-chain`**.

`robinhood` would have matched no chain and 404ed silently, after being baked in by `initialize(string)` — exactly the failure mode you flagged. The value now carries its derivation in the plan doc so the next reader doesn't re-raise it.

**On the `deploy_17` guard** — agreed the config can't enforce it, since `deploy_17_service_staking_native_token.sh` doesn't read `stakingNativeTokenAddress` at all. I've left the JSON alone rather than adding a pseudo-key: a comment in a file consumed by `jq -r '.key'` isn't a guard either, and a real one belongs in the script. The rollout plan's §3.2 step list omits `deploy_17` and §10 R1 says why. Happy to add a guard to the script itself in a separate PR if you'd prefer it enforced rather than documented.

bennyjo (approved):
Re-approving at d91e33f. Checksum fixed, and the slug change back to robinhood-chain matches the frontend's kebab-case rule against viem's chain name for 4663.

PR #325, opened 2026-09-09
deploy(robinhood): wave 1 on chain 4663, and four shell-route script fixes

DavidMinarsch:
Wave 1 of the Robinhood Chain (4663) rollout: 15 contracts deployed and wired, plus **six defects found in the shell deployment route** — all pre-existing on `main`, all affecting every chain deployed that way. 4663 is also **registered for the fleet audit** here, which is what both reviewers asked for.

Deployed from `0xEB2A22b27C7Ad5eeE424Fd90b376c745E60f914E`, 30 transactions, ~0.0064 ETH. **Ownership is deliberately still with the deployer** — the handover to the bridge mediator waits on this review.

**Eleven files: the globals, six script fixes, two preflight guards, the audit registration (two files), and two `.gitleaksignore` fingerprints.** No build artifacts.

## Deployed

| contract | address |
|---|---|
| `ServiceRegistryL2` | `0xE3607b00E75f6405248323A9417ff6b39B244b50` |
| `OperatorWhitelist` | `0x3C1fF68f5aa342D296d4DEe4Bb1cACCA912D95fE` |
| `ServiceRegistryTokenUtility` | `0x3d77596beb0f130a4415df3D2D8232B3d3D31e44` |
| `ServiceManager` (impl) | `0x34C895f302D0b5cf52ec0Edd3945321EB0f83dd5` |
| **`ServiceManagerProxy`** | **`0x63e66d7ad413C01A7b49C7FF4e3Bb765C4E4bd1b`** ← the live one |
| `GnosisSafeMultisig` | `0xBb7e1D6Cb6F243D6bdE81CE92a9f2aFF7Fbe7eac` |
| `GnosisSafeSameAddressMultisig` | `0xFbBEc0C8b13B38a9aC0499694A69a10204c5E2aB` |
| `RecoveryModule` | `0xE43d4F4103b623B61E095E8bEA34e1bc8979e168` |
| `SafeMultisigWithRecoveryModule` | `0xb09CcF0Dbf0C178806Aaee28956c74bd66d21f73` |
| `ComplementaryServiceMetadata` | `0xD1155408D58293BE0743225bcDe28b9FD0C12378` |
| `IdentityRegistryBridger` (impl) | `0xa4799B083E0068732456EF45ff9fe5c683658327` ← **redeployed, see below** |
| **`IdentityRegistryBridgerProxy`** | **`0xE49CB081e8d96920C38aA7AB90cb0294ab4Bc8EA`** ← the live one, address unchanged |
| `StakingToken` (**0.2.0**) | `0x87c511c8aE3fAF0063b3F3CF9C6ab96c4AA5C60c` |
| `StakingVerifier` | `0x75D529FAe220bC8db714F0202193726b46881B76` |
| `StakingFactory` | `0x1BD1505B711Fb58C54ca3712e6BEf47A133892d9` |

Every constructor argument was read back on-chain, and every contract's runtime bytecode compared against its live counterpart on another chain — each difference accounted for as an inlined immutable or `address(this)`, with nothing unexplained.

## The fleet audit now covers 4663

Both reviewers made this the merge condition, and both were right that it is two files, not one: `docs/configuration.json` **and** the hardcoded `globalNames` / `providerLinks` maps in `audit_contracts_setup.js`. JSON alone throws on `readFileSync(undefined)`.

The fifteen artifact paths in the configuration entry were not chosen by name. Each deployed runtime was compared against every committed artifact with the CBOR metadata trailer stripped: four match byte-for-byte, the rest differ only in inlined immutables.

**The audit run this enables was not clean, and that is the point.** It found two wiring defects that the one-off read-back had missed — see the next section. After the fix it reports, on 4663:

- six `owner()` still the deployer rather than the bridge mediator — expected, that is the pending handover, exactly as OjusWiZard predicted (`ownershipChangeRequired: yes`)
- both `drainer()` were zero — **fixed here**, see below
- metadata-trailer WARNs and `HashCheckpoint: no configuration entry`, both of which appear identically on Arbitrum, Optimism and Base

**The final run is clean apart from those six `owner()` lines** — no other assertion fails on 4663, and the process exits 0 with no bytecode-length mismatch.

`ownable_owners.csv` is deliberately not regenerated here: it is a generated file, already stale on `main` for other chains, and every 4663 row flips at handover. It belongs in the handover PR.

## The 8004 bridger was dead on arrival — found by the audit, fixed here

`IdentityRegistryBridger` derives its manager rather than being told it:

```solidity
address public immutable serviceManager;                          // :100
serviceManager = IServiceRegistry(_serviceRegistry).manager();    // :135
```

`deploy_27` ran before `script_l2_02_change_managers_registries.sh`, so the constructor read zero and baked it in. Every `register` / `updateAgentWallet` path would have reverted `ManagerOnly`. Separately, `ServiceManagerProxy.setIdentityRegistryBridger` had never been called, so the binding was missing in both directions.

**No setter can repair it.** There is no `changeServiceManager`; the only mutable state is `owner`, `baseURI`, `startLinkServiceId` and two mappings. And immutables are inlined into the *implementation's* runtime code, so the proxy's `delegatecall` reads the same zero — storage cannot override it. The bytecode says so independently: diffing the deployed implementation against a fresh build shows 236 bytes over 16 runs, 8 sites inlining `identityRegistry` and 5 inlining `serviceRegistry`, and **zero sites for `serviceManager`**, i.e. its inlined value equalled the compiler's zero placeholder.

Fixed with two scripts already in the repo — `deploy_27` again, then `script_l2_06` — plus `script_l2_05` for the binding. **The proxy keeps its address and all storage**; `owner`, the ERC-8004 base URI, `startLinkServiceId = 1` and both mappings were re-read afterwards and are intact. `totalSupply()` is 0, so there was no linked service state to migrate.

Now reads `IdentityRegistryBridgerProxy.serviceManager() = 0x63e6…bd1b` and `ServiceManagerProxy.identityRegistryBridger() = 0xE49C…c8EA`.

**Why it happened, since the ordering constraint is written down nowhere.** `docs/deploymentL2.md` stops at step 15 — it is the 2023-era light-protocol sequence and covers none of 24–28. The L2 README's "run in strict numerical order" rule addresses the hardhat `deploy_NN_*.js` family, which has no 24–28 at all; those five exist only as `.sh`. In the shell world the dependency crosses from `deploy_27_*` to `script_l2_02_*` — two independently numbered sets — so running every `deploy_*` and then every `script_l2_*` gets it backwards. It only bites a chain built in one pass; everywhere else the registry had a live manager long before the 8004 contracts were added.

**Guarded, so it cannot recur.** `deploy_27` now reads `manager()` on-chain before `forge create` and refuses to deploy against a zero. `ComplementaryServiceMetadata` is the same shape — it copies `baseURI` out of the registry in its constructor and has no `changeBaseURI` — so `deploy_26` got the matching guard. Both exit 1, not the `deploy_*` family's legacy `exit 0` that a caller without `set -e` would run straight past. Exercised against three fixtures on 4663: the live registry passes both; the abandoned registry `0x9338b515…`, whose `manager()` is genuinely unset, trips the `deploy_27` guard; an address with no code trips both.

A sweep of all fifteen deployed contracts for the same failure class — any constructor making an external read — turns up only these two. The other thirteen read nothing.

## Slashed-fund drainers set

Both registries had `drainer() == 0`, which the audit surfaced. `drainer` is the destination for slashed operator bonds — `ServiceRegistryL2.slashedFunds` (native) and `ServiceRegistryTokenUtility.mapSlashedFunds[token]` (ERC-20) — and with it unset both sweep paths are closed: `ServiceRegistryL2.drain()` requires `msg.sender == drainer`, and `ServiceRegistryTokenUtility.drain(token)` reverts `ZeroAddress()` outright.

Nothing was at risk — `totalSupply()` is 0, so no service exists and no operator is bonded. It had to happen **before** the handover only because `changeDrainer` is owner-gated, and afterwards it would cost a governance proposal crossing the bridge instead of one `cast send`.

| tx | block | contract |
|---|---|---|
| `0x56fb160340c2a9cc0bddeadd5154054a795900c2b9bfdf8e6606901449f99304` | 58763774 | `ServiceRegistryL2` |
| `0x95512f8f2512ea87913e9d875592c2ed3cba39589dc1e7db9d7b3a2559d20363` | 58764035 | `ServiceRegistryTokenUtility` |

Both now read `0x4d30F68F5AA342d296d4deE4bB1Cacca912dA70F`, the same bridge-mediator convention Base, Gnosis and Celo use on both contracts. These are steps 11–12 of `deploymentL2.md`, which precede the ownership transfer in steps 13–15.

## Six script fixes

**1. `deploy_01` passed the wrong constructor arguments.** `serviceRegistryName` is `"Service Registry L2"`, and the script built the whole `forge` invocation as a string then ran `$($execCmd)`, so it re-split on expansion. The first deployment came out with `name() = "Service"`, `symbol() = "Registry"`, `baseURI() = "L2"`. Quoting the variable alone would not fix it; the fix calls forge directly with each argument as its own quoted word.

> **Correction, and thanks to @OjusWiZard for catching it.** The description previously said forge's arity check could not catch this "because the count still comes to three". That was wrong: the split produced **five** words, not three, and forge consumes the first three and **silently drops the surplus** rather than rejecting the count. I ran it to be sure — forge 1.5.1 on anvil, five words into the three-string constructor, deploys successfully and reproduces `"Service" / "Registry" / "L2"` exactly. That closes the caveat OjusWiZard flagged on their own inference, and it means over-splitting is silent whenever the consumed prefix type-checks, irrespective of count.

**2. That same defect is still armed on L1.** Following the corrected mechanism through the L1 scripts: `deploy_01_component_registry.sh` has `componentRegistryName = "Component Registry"` — four words into three string parameters, all string, so it is the **silent** form. Deployed on anvil from the committed globals it yields `name() = "Component"`, `symbol() = "Registry"`, `baseURI() = "AUTONOLAS-COMPONENT-V1"`. Fixed the same way. `deploy_02` and `deploy_04` take an `address` fourth, so the same mis-split fails loudly — left alone. **Nothing has shipped through any of them**: live `ComponentRegistry`, `AgentRegistry` and `ServiceRegistry` on Ethereum all read correctly, because L1 went the hardhat route. The defect was latent with its trigger data committed.

**3. `deploy_06a` passed the `initialize()` selector as the Safe proxy factory.** Line 32 read `gnosisSafeProxyFactoryAddress=$(cast calldata "initialize()")` — a copy-paste of `deploy_05`'s legitimate `proxyData` line. It failed loudly here because `0x8129fc1c` is not address-shaped. Had it been, it would have been baked in: `GnosisSafeMultisig` holds the factory as an `immutable` and creates every service's Safe.

**4. `deploy_06a`/`06b` wrote globals keys nothing reads.** They wrote `gnosisSafeMultisigAddress` / `gnosisSafeSameAddressMultisigAddress` while the globals and the hardhat reader use the `…ImplementationAddress` spelling — leaving a populated stray key beside an empty real one. Same writer/reader split as the `dispenserProxyAddress` defect in valory-xyz/autonolas-tokenomics#358. Normalised.

**5. `script_l2_01` never whitelisted the two Gnosis Safe multisigs.** It covered only `SafeMultisigWithRecoveryModule` and `RecoveryModule`, so a chain deployed via the shell route alone ends up with `GnosisSafeMultisig` and `GnosisSafeSameAddressMultisig` deployed but **not permitted**, and `ServiceRegistry.deploy()` reverts `UnauthorizedMultisig` for any service using them. Only the hardhat route whitelisted them — and that route is excluded on 4663 because Etherscan does not cover the chain.

**6. `script_l2_01` could not tell success from failure.** Per @OjusWiZard: the `cast send` calls checked the result with a bare `grep "status"`, which matches `status 0 (failed)` as readily as `status 1 (success)`. All four now use the `script_l2_07` form and `exit 1` on failure — not just the two this PR added, since leaving two weak checks beside two strong ones in the same file is worse.

All six are in the shell route, none in the hardhat route. Worth noting as a pattern: the shell path is newer and less exercised, and this rollout is committed to it.

**Accepted, not done here:** `deploy_06a`/`06b` still build `$execCmd` as a string — same hazard class, safe today because addresses and a `bytes32` cannot contain spaces. Left for the follow-up @dagacha suggested.

## StakingToken is pinned to 0.2.0 — deliberately

`deploy_16` builds from `main`, which is now **0.3.0** (`RewardDistributionType`, commit `067990b`). Every other chain runs **0.2.0**. Deploying 0.3.0 here would have made 4663 the only chain running a build the rest of the fleet does not. The decision was to keep the fleet uniform and move every chain to 0.3.0 together, as a deliberate fleet-wide step rather than a side effect of a new chain launch.

So it was built from commit `5efeb59` with solc 0.8.28 / 750 runs / cancun — that commit's hardhat settings, since the repo had no `foundry.toml` then — and reproduces Mode's live implementation exactly once metadata is stripped.

⚠️ **`deploy_16` still has no version pin**, so re-running it produces 0.3.0. Not fixed here because it is a fleet-wide decision, not a Robinhood one.

## Verified state

- both registries' `manager()` → `ServiceManagerProxy`
- proxy's `operatorWhitelist()` → `OperatorWhitelist`
- `ServiceManagerProxy.identityRegistryBridger()` → bridger proxy, and the bridger proxy's `serviceManager()` → `ServiceManagerProxy` — the mutual binding, closed in this PR
- all four multisig implementations allowlisted; an arbitrary control address is not
- `StakingVerifier.implementationsCheck() == true`, `StakingToken` 0.2.0 allowlisted, the abandoned 0.3.0 build **not** allowlisted, control **not** allowlisted
- **no `StakingNativeToken` exists on this chain and none is allowlisted** — the Mode G4 mechanism cannot arise here
- both `drainer()` → the bridge mediator, matching Base, Gnosis and Celo on both contracts
- all `owner()` still the deployer

## Three abandoned contracts, referenced by nothing

`0x9338b515…` (the mis-quoted `ServiceRegistryL2`), `0x88DE7346…` (the 0.3.0 `StakingToken`), and now `0x397125902ED2cA2d42104F621f448A2cE1bC8Fb7` (the bridger implementation with the zero `serviceManager`). The first two are worth knowing about because those addresses collide with meaningful contracts elsewhere through deployer-nonce reuse — `0x9338b515…` is Gnosis's `ServiceRegistryL2`, Polygon's `FxGovernorTunnel` and Mode's mediator; `0x88DE7346…` is Mode's superseded V1 `StakingNativeToken` that proposal 13 just de-allowlisted.

## Explorer verification — happening on its own

Etherscan does not list 4663; Robinhood's Blockscout API sits behind a Cloudflare challenge that 403s non-browser clients (a Blockscout Pro key proxies reads fine, but its verification write path 500s and `verification/config` 404s); Sourcify supports the chain but its single 4663 RPC was failing throughout the deployment.

None of that mattered in the end. **Blockscout is auto-verifying via its cross-chain bytecode database** (`is_verified_via_eth_bytecode_db: true`) — with almost no manual submission. That works precisely because every contract has a byte-identical twin on another chain once immutables are accounted for, which is the same property the bytecode comparisons above establish.

Build trees, should anything need reproducing:

| contracts | source | solc | optimizer | evmVersion |
|---|---|---|---|---|
| all except `StakingToken` | `main` @ `52ee868` | 0.8.30 | on, 750 runs | prague |
| **`StakingToken`** | **commit `5efeb59`** | **0.8.28** | on, 750 runs | **cancun** |

The redeployed bridger uses the first row, and its fresh build is **byte-identical** to the committed `abis/0.8.30/IdentityRegistryBridger.json`.

## Next

Handover of `owner()` to the bridge mediator `0x4d30F68F5AA342d296d4deE4bB1Cacca912dA70F` after this review — the last thing outstanding.

Note `changeOwner` on `ServiceRegistryL2` and `ServiceRegistryTokenUtility` has **no shell script**; both are covered only by the hardhat `deploy_13_15`, so they need direct `cast send`. The same was true of `changeDrainer` (hardhat `deploy_11_12` only), which is why the two calls below were sent directly.

⚠️ **Please merge with a merge commit**, per @dagacha — the `.gitleaksignore` fingerprints are commit-SHA-pinned and would break under squash or rebase.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


DavidMinarsch:
Dropped the `verification/` folder — it was 440K of build artifacts across 32 files, and it was what gitleaks flagged (hex addresses inside the standard-JSON, same false-positive class as the `.gitleaksignore` entries in the marketplace repo).

Nothing is lost. The build trees are recorded in the PR body, and regenerating is one `forge verify-contract --show-standard-json-input` per contract:

| contracts | source | solc | optimizer | evmVersion |
|---|---|---|---|---|
| all except `StakingToken` | `main` @ `52ee868` | 0.8.30 | on, 750 runs | prague |
| **`StakingToken`** | **commit `5efeb59`** | **0.8.28** | on, 750 runs | **cancun** |

**As it turns out they weren't needed.** Blockscout is auto-verifying the deployment through its cross-chain bytecode database (`is_verified_via_eth_bytecode_db: true`) — **11 of 15 already**, with no manual submission for most of them. That works precisely because every contract has a byte-identical twin on another chain once immutables are accounted for, which is the same property the PR's bytecode comparisons establish.

The PR is now five files: the four script fixes and the globals.

scripts/deployment/l2/globals_robinhood_mainnet.json
```diff
@@ -25,20 +25,20 @@
...
-  "serviceManagerProxyAddress": "",
-  "gnosisSafeMultisigImplementationAddress": "",
-  "gnosisSafeSameAddressMultisigImplementationAddress": "",
-  "recoveryModuleAddress": "",
-  "safeMultisigWithRecoveryModuleAddress": "",
-  "complementaryServiceMetadataAddress": "",
-  "identityRegistryBridgerAddress": "",
-  "identityRegistryBridgerProxyAddress": "",
-  "stakingTokenAddress": "",
+  "serviceRegistryAddress": "0xE3607b00E75f6405248323A9417ff6b39B244b50",
```
OjusWiZard, line 28: **The chain this PR deploys is not registered for the fleet audit.**

`docs/configuration.json` lists 8 chains — 1, 137, 100, 42161, 10, 8453, 42220, 34443 — and no 4663. `audit_chains/audit_contracts_setup.js` iterates that file, so an absent chain is not "unaudited and flagged"; it is invisible, and the process still exits 0.

What that costs specifically: the audit's `mapMultisigs` check is the standing, repeatable version of the very thing fix 4 repairs. Same for constructor arguments and `manager()` wiring — the audit covers exactly the defect class all four fixes address. Right now 4663's only evidence is the one-off manual read-back in this PR description, and a one-off manual read-back is what produced these four defects in the first place.

I verified the deployed state myself and it is correct today, so this is not about whether wave 1 landed properly — it did. It is that the chain has no continuous check, at the moment it acquires 15 contracts whose ownership is about to move to the bridge mediator, after which corrections cost a governance cycle.

Adding the 4663 entry here would also give the audit something to compare against, since this file is now the canonical record of those addresses.

If registration is deliberately deferred to wave 2 — plausible, since ownership is still with the deployer — worth saying so explicitly, so it reads as sequenced rather than forgotten.

OjusWiZard (dismissed):
Approving. I verified the deployment against chain 4663 directly rather than reading the table, and every claim holds.

### Verified on-chain

- **All 15 contracts have code.** Wiring: both registries' `manager()` → `ServiceManagerProxy` `0x63e66d7a…`, and the proxy's `operatorWhitelist()` → `OperatorWhitelist` `0x3C1fF68f…`.
- **Multisig allowlist:** all four implementations `true`, a control address `false`.
- **StakingVerifier:** `implementationsCheck()` true, 0.2.0 allowlisted, the abandoned 0.3.0 **not** allowlisted, control not allowlisted, `StakingFactory.verifier()` correct.
- **Ownership:** every `owner()` still the deployer, as stated. (`ComplementaryServiceMetadata` has no `owner()` at all — so its absence from the handover list is correct, not an omission.)
- **The 0.2.0 pin checks out.** `StakingToken` on 4663 vs Mode's: 13,898 bytes raw on both, **byte-identical** after stripping the 53-byte metadata trailer. It really is the fleet's 0.2.0, not a lookalike rebuild.
- **Defect 1 is demonstrable.** The abandoned `0x9338b515…` reports `name()='Service'`, `symbol()='Registry'`, `baseURI()='L2'`; the live one reports the correct three.

### One correction to the description, because it changes what else is at risk

> forge's arity check cannot catch this because the count still comes to three.

The count came to **five**, not three: `Service | Registry | L2 | AUTONOLAS-SERVICE-L2-V1 | https://gateway.autonolas.tech/ipfs/`. The abandoned contract holds exactly words 1-3, so `forge create --constructor-args` **tolerated the surplus and dropped the rest**. Not a coincidental arity match.

That matters for the sibling scripts, because it means over-splitting is silent whenever the consumed prefix type-checks — regardless of count. Re-deriving the three L1 scripts on that basis, against the values already committed in `globals_mainnet.json`:

| script | words → params | outcome |
|---|---|---|
| `deployment/deploy_01_component_registry.sh` | 4 → 3, all `string` | **silent** — drops `baseURI`, gives `name()='Component'`, `symbol()='Registry'`, `baseURI()='AUTONOLAS-COMPONENT-V1'` |
| `deployment/deploy_02_agent_registry.sh` | 5 → 4, 4th is `address` | loud — a URL lands in the address slot |
| `deployment/deploy_04_service_registry.sh` | 5 → 4, 4th is `address` | loud — same |

So one L1 script carries the silent form, not all three. **Nothing has shipped**: live `ComponentRegistry`, `AgentRegistry` and `ServiceRegistry` on Ethereum all read correctly, so L1 went through the hardhat route. The defect is latent, with the trigger data committed, on the route this rollout is committed to.

Caveat on my own claim: I could not run `forge` here, so the surplus-tolerance mechanism is inferred from the deployment artifact rather than confirmed against forge. Five words in and three stored is strong evidence, but worth one deliberate-surplus run to nail down before relying on it.

### One inline comment

On the audit-registration gap — see below. Separately and not blocking: the two new `cast send` calls in `script_l2_01` only `grep "status"`, which matches `status 0 (failed)` as readily as `status 1 (success)`; `script_l2_07:115-118` in the same directory already has the check that distinguishes them, if you want it while the file is open.

dagacha:
One ask before merge.

**Say what happens to fleet audit — in the PR body, not as a later thought.** Either register 4663 now, or write that registration is wave 2 / valory-xyz/autonolas-analytics#254.

Silent omission is not “unaudited”. `docs/configuration.json` has no 4663, so `audit_chains/audit_contracts_setup.js` never sees the chain and still exits 0. The audit’s `mapMultisigs` / `manager()` / constructor checks are the standing version of the four shell-route defects this PR just fixed. Right now 4663’s only evidence is a one-off read-back.

If you do add it, it is two files, not one: `configuration.json` **and** the hardcoded `globalNames` / `providerLinks` maps in `audit_contracts_setup.js`. JSON-only throws on `readFileSync(undefined)`. `checkOwner` on L2 will then log deployer ≠ `bridgeMediatorAddress` (`ownershipChangeRequired: yes`) until handover; that is expected and useful.

I verified the deployed state on RPC `https://rpc.mainnet.chain.robinhood.com` (chain id 4663) independently of the table. Wave 1 is correct; the four script fixes are correct. This is the only merge demand.

---
Reviewed with [omp](https://github.com/oh-my-pi/omp) harness. Main LLM: grok-4.6. Effort: not labeled in-session (independent on-chain read-back, not a skim of the description).

dagacha (approved):
**Approval — verified independently, not just read.**

Review performed with **ZCode** (autonomous coding agent), powered by **GLM-5.3** (Zhipu), at **high effort**: full diff audit of all six files plus independent on-chain verification of the deployed state via the chain's public RPC.

## Verified on chain 4663 (independent read-backs via `https://rpc.mainnet.chain.robinhood.com`)

- `ServiceRegistryL2` reads back `name() = "Service Registry L2"`, `symbol() = "AUTONOLAS-SERVICE-L2-V1"`, `baseURI() = "https://gateway.autonolas.tech/ipfs/"` — the deploy_01 fix is live in the actual deployment. The abandoned first attempt at `0x9338b515…` reads back `name() = "Service"`, `symbol() = "Registry"`, exactly the mis-quote described.
- Both registries' `manager()` → `ServiceManagerProxy (0x63e6…bd1b)`; the proxy's `operatorWhitelist()` → `OperatorWhitelist (0x3C1f…95fE)`; `VERSION()` through the proxy = `1.2.0`; bridger proxy `baseURI()` = the `robinhood-chain` ERC-8004 slug.
- All four multisig implementations return `mapMultisigs() == true`; a control address returns `false` — the script_l2_01 whitelist additions were executed.
- `StakingVerifier.implementationsCheck() == true`; StakingToken 0.2.0 (`0x87c5…C60c`) allowlisted; the abandoned 0.3.0 build (`0x88DE…3728`) and a control both rejected.
- `owner()` on the registry, proxy, and verifier is still the deployer `0xEB2A…914E` — consistent with the handover pending this review.
- **Bytecode twins**: `GnosisSafeMultisig`, `GnosisSafeSameAddressMultisig`, and `StakingToken` are byte-identical to Mode's live contracts after stripping metadata — the StakingToken match confirms the 0.2.0 pin (commit `5efeb59`, solc 0.8.28 / 750 runs / cancun) reproduced Mode's build exactly.
- All 17 addresses in the PR (15 deployed + deployer + bridge mediator) pass EIP-55 checksum validation.

## Verified in the diff

- **deploy_01**: direct `forge create` with each constructor arg as its own quoted word is the correct fix — re-quoting `$execCmd` alone would not have helped, since the re-split happens at expansion. `$walletArgs` correctly left unquoted (multi-word).
- **deploy_06a/06b key rename**: the `…ImplementationAddress` spelling matches every reader (`audit_chains/audit_contracts_setup.js:335`, `deploy_07_10_change_managers_and_permissions.js:18-19`, `deploy_10_14_change_managers.js:20-21`, proposals 12/13). No committed globals file carries the old stray keys. The `gnosisSafeProxyFactoryAddress` read is correct (canonical `0xa6B7…6AB2`).
- **script_l2_01**: the two appended whitelist calls follow the file's existing pattern.
- **.gitleaksignore**: local gitleaks 8.30.1 scan of the branch is clean (1315 commits, no leaks); both fingerprints pin real commits on the branch and the pinned lines match the flagged keys (`0c119407` → line 30 = `serviceRegistryTokenUtilityAddress`; `264765d0` → line 40 = `stakingTokenAddress`).

## Non-blocking notes

1. `deploy_06a`/`06b` still deploy via the string-built `$execCmd` pattern — same hazard class as the deploy_01 bug. Safe today because their constructor args (two addresses / a bytes32) cannot contain spaces, but worth normalizing in a follow-up.
2. The gitleaks fingerprints are commit-SHA-pinned: they survive this repo's merge-commit style but would break under a squash/rebase merge — please merge with a merge commit.
3. `script_l2_01` continues past a failed `cast send` (exit 0 either way) — pre-existing pattern throughout the file, not introduced here.
4. Deployer-nonce reuse makes cross-chain address collisions a real reviewing trap (e.g. `0x3C1fF68…` is Mode's `ServiceRegistryL2` but Robinhood's `operatorWhitelist`/`timelock`, matching Celo's globals) — the PR flags this and it checks out.

dagacha (dismissed):

OjusWiZard (dismissed):

scripts/deployment/l2/globals_robinhood_mainnet.json
DavidMinarsch: Done — and you were right that it is two files, not one. `docs/configuration.json` plus the hardcoded `globalNames` / `providerLinks` maps in `audit_contracts_setup.js` (`0c8e4af`). RPC `https://rpc.mainnet.chain.robinhood.com`.

Not deferred to wave 2, because your reasoning about *why* it matters turned out to be load-bearing:

**The first audit run was not clean.** It found two wiring defects the one-off read-back had missed — exactly your point that a one-off read-back is what produced these defects in the first place.

`IdentityRegistryBridger` had `serviceManager == 0`. It is an `immutable` assigned in the constructor from `IServiceRegistry(_serviceRegistry).manager()`, and `deploy_27` ran before `script_l2_02_change_managers_registries.sh`. No setter exists, and an immutable lives in the implementation runtime code, so the proxy's `delegatecall` reads the same zero — a redeploy was the only repair. Every `register` / `updateAgentWallet` path would have reverted `ManagerOnly`. Separately `ServiceManagerProxy.setIdentityRegistryBridger` had never been called, so the binding was missing in both directions. Both fixed; the proxy kept its address and all storage.

Your `checkOwner` prediction holds exactly: six `owner()` log deployer ≠ `bridgeMediatorAddress` until handover. The audit also surfaced both `drainer()` still zero, which was genuinely outstanding and now goes with the handover batch.

On the artifact paths in the configuration entry — I did not pick them by name. Each deployed runtime was compared against every committed artifact with the CBOR trailer stripped; four match byte-for-byte, the rest differ only in inlined immutables.

**Your correction to the description was right, and I ran it rather than leaving it inferred.** forge 1.5.1 on anvil, five words into `ServiceRegistryL2`'s three-string constructor: deploys successfully, reproduces `"Service" / "Registry" / "L2"`. So it is surplus-tolerance, not an arity match, and your caveat about not being able to confirm it against forge is closed.

Following that through: `deployment/deploy_01_component_registry.sh` is the silent one, as you derived. On anvil from the committed globals it gives `name()="Component"`, `symbol()="Registry"`, `baseURI()="AUTONOLAS-COMPONENT-V1"`. Fixed in `610896d`. `deploy_02` / `deploy_04` fail loudly on the address parameter, so they are left alone.

Also took the `grep "status"` point — all four calls in `script_l2_01` now use the `script_l2_07` form and `exit 1`, not just the two this PR added.

And `deploy_27` / `deploy_26` now preflight the constructor reads on-chain and refuse to deploy against a zero, so this particular ordering trap cannot recur.

DavidMinarsch:
@dagacha — done, and stated in the PR body rather than as a later thought.

**4663 is registered now, not deferred** (`0c8e4af`): `docs/configuration.json` and the `globalNames` / `providerLinks` maps in `audit_contracts_setup.js`, exactly the two files you and @OjusWiZard both named.

Your reason for demanding it was correct in a way I did not expect — **the first audit run failed**, on two wiring defects the one-off read-back had missed:

1. `IdentityRegistryBridger.serviceManager() == 0`. It is an `immutable` read in the constructor from `IServiceRegistry(_serviceRegistry).manager()`, and `deploy_27` ran before `script_l2_02_change_managers_registries.sh`. No setter, and an immutable lives in the implementation runtime code so the proxy reads the same zero — redeploy was the only repair. Every `register` / `updateAgentWallet` would have reverted `ManagerOnly`.
2. `ServiceManagerProxy.setIdentityRegistryBridger` had never been called.

Both fixed in `610896d`; the proxy `0xE49CB081…` kept its address and all storage. The audit also surfaced both `drainer()` still zero — genuinely outstanding, now sequenced with the handover batch.

`deploy_27` and `deploy_26` now preflight those constructor reads on-chain and `exit 1` rather than deploying against a zero.

**On your two non-blocking notes:**

`deploy_06a`/`06b` `$execCmd` — accepted as a follow-up, per your framing, and said so in the body. Safe today since addresses and a `bytes32` cannot contain spaces.

Merge commit — noted in the PR body with a ⚠️, and you are right about the fingerprints. Worth adding that `610896d` changes `identityRegistryBridgerAddress` **in place**, so the two pinned lines (30 and 40) do not shift and both existing fingerprints stay valid.

One correction to the description worth flagging since you verified against it: @OjusWiZard caught that the deploy_01 mis-split produced **five** words, not three, and that forge silently drops the surplus rather than rejecting the count. I confirmed it on anvil with forge 1.5.1. That generalises the defect — over-splitting is silent whenever the consumed prefix type-checks — and it turns up one more instance, `deployment/deploy_01_component_registry.sh` on L1, fixed here. Nothing shipped through it; L1 went the hardhat route.

DavidMinarsch:
posted

docs/configuration.json
```diff
@@ -686,5 +686,86 @@
...
+      },
+      {
+        "name": "ComplementaryServiceMetadata",
+        "artifact": "abis/0.8.30/ComplementaryServiceMetadata.json",
+        "address": "0xD1155408D58293BE0743225bcDe28b9FD0C12378"
+      },
+      {
+        "name": "IdentityRegistryBridger",
+        "artifact": "abis/0.8.30/IdentityRegistryBridger.json",
+        "address": "0x397125902ED2cA2d42104F621f448A2cE1bC8Fb7"
```
OjusWiZard, line 762: This is the abandoned `IdentityRegistryBridger` implementation — the one with `serviceManager() == address(0)` that `610896d1` redeployed to fix. Should be `0xa4799B083E0068732456EF45ff9fe5c683658327`.

The live system is fine; only the published registry is wrong:

```
proxy 0xE49CB081... impl slot (keccak "PROXY_IDENTITY_REGISTRY_BRIDGER")
                         -> 0xa4799B083E0068732456EF45ff9fe5c683658327
globals_robinhood_mainnet.json:38
                            0xa4799B08...   agrees with chain
this line                   0x39712590...   does not

0x39712590...  7861 bytes  serviceManager() -> 0x0000...0000
0xa4799B08...  7861 bytes  serviceManager() -> 0x63e66d7a...
proxy           304 bytes  serviceManager() -> 0x63e66d7a...
```

Every other chain's non-`Proxy` entry matches its globals `identityRegistryBridgerAddress` (42161, 8453, 42220, 100, 10, 137), so 4663 is the only one out of step.

Worth fixing beyond the docs value: `checkOwner` and `recordOwnershipRow` also run against whatever this entry names, so the ownership CSV currently records 4663's bridger owner from the dead implementation.

OjusWiZard (approved):
Approving. The bridger repair verifies: I read all 16 `immutable`s across the 4663 deployment on-chain and every one is correct, including the new bridger's three and `GnosisSafeSameAddressMultisig.proxyHash` matching `multisigProxyHash130` byte-for-byte.

One inline finding on `docs/configuration.json` — it's a one-line fix and the chain itself is already correct, so it doesn't block.

---

Separately, correcting something I said in my earlier round on this PR. I described registering 4663 in `audit_contracts_setup.js` as making the audit cover Robinhood. That's weaker than I stated: `bytecodeMismatchFound` is the only thing `process.exit(1)` reads, and it's set at exactly two sites, both bytecode-length checks. All 58 `customExpect` wiring assertions print on failure and cannot move the exit code.

That matters for the finding below, because the assertion written for exactly this defect is already in the file:

```js
const serviceManager = await identityRegistryBridger.serviceManager();
customExpect(serviceManager, globalsInstance["serviceManagerProxyAddress"], ...);
```

`findContractInstance` resolves `IdentityRegistryBridger` by name straight out of `configuration.json`, so the audit does dereference the stale address and does read `address(0)` back. It will print the mismatch and exit 0.

The bytecode tier can't catch it either: both implementations are 7861 bytes (same contract, different constructor args — immutables differ, length doesn't), so Tier 1 passes, and Tier 2 compares the CBOR trailer, which is identical for the same source and compiler.

Not asking for anything here — the audit's exit-code semantics predate this PR. Flagging it because "the audit now covers 4663" is worth reading as "the audit now prints about 4663".

docs/configuration.json
DavidMinarsch: Confirmed and fixed in #326.

You are right on both counts, including the second-order one. `0c8e4af` recorded the implementation address, then `610896d` redeployed it and wrote the new address to the globals only — I never updated `configuration.json`, so the published registry kept pointing at the implementation the redeploy existed to replace.

Verified against merged main before fixing: 4663 was the only chain out of step. 42161, 8453, 42220, 100, 10 and 137 all agree with their globals.

The ownership-CSV consequence you flagged is the part worth having caught — `checkOwner` and `recordOwnershipRow` run against whatever the entry names, so the recorded bridger owner for 4663 came from the dead implementation rather than the live one.

#326 also closes the two items #325 deferred: the `deploy_06a`/`06b` $execCmd normalisation @dagacha suggested, and shell scripts for `changeDrainer`/`changeOwner` on `ServiceRegistryL2` and `ServiceRegistryTokenUtility` — the two contracts the shell route never covered, which is why those calls were sent by hand on 4663.

PR #326, opened 2026-09-09
fix(robinhood): correct the 4663 bridger entry, and the two follow-ups #325 left open

DavidMinarsch:
Follow-up to #325 (merged). One defect it shipped, plus the two items it explicitly deferred.

## 1. `configuration.json` names the dead bridger implementation

Caught by @OjusWiZard on the merged PR. `0c8e4af` recorded the implementation address, then `610896d` redeployed it to repair its zero `serviceManager` immutable and wrote the new address **to the globals only** - `configuration.json` was never updated, so the two disagree.

```
proxy 0xE49CB081...c8EA  getImplementation() -> 0xa4799B083E0068732456EF45ff9fe5c683658327
globals_robinhood_mainnet.json                  0xa4799B083E0068732456EF45ff9fe5c683658327
configuration.json                              0x397125902ED2cA2d42104F621f448A2cE1bC8Fb7  <- dead

0x39712590...  serviceManager() -> 0x0000...0000   (the implementation 610896d replaced)
0xa4799B08...  serviceManager() -> 0x63e66d7a...   (live)
```

**4663 was the only chain out of step** - 42161, 8453, 42220, 100, 10 and 137 all agree with their globals.

The live system was never affected. But `checkOwner` and `recordOwnershipRow` run against whatever this entry names, so the ownership CSV was recording 4663's bridger owner from the dead implementation.

## 2. Two scripts the shell route never had

`ServiceRegistryL2` and `ServiceRegistryTokenUtility` are reachable only through the hardhat `deploy_11_12` / `deploy_13_15`. On a shell-route chain there is no script at all - which is why both `changeDrainer` calls on 4663 had to be sent by hand, and why `changeOwner` on those two still has no scripted path for the pending handover.

| script | call |
|---|---|
| `script_l2_09_registries_change_drainer.sh` | `changeDrainer(bridgeMediator)` on both |
| `script_l2_10_registries_change_owner.sh` | `changeOwner(bridgeMediator)` on both |

**Order matters and the scripts say so:** `_09` must run before `_10`, because `changeDrainer` is owner-gated - once `owner()` is the bridge mediator it becomes a governance proposal crossing the bridge rather than one transaction. That is why `deploymentL2.md` puts the drainers at steps 11-12 and ownership at 13-15.

Both follow `script_l2_07`'s existing shape: idempotent per contract, signer pre-checked against `owner()` before attempting, `status 1` verified, `exit 1` on failure.

**Exercised against 4663**, where both drainers are already set - `script_l2_09` correctly reports both as no-ops and exits 0:

```
ServiceRegistryL2 0xE3607b00...4b50 already has drainer() == 0x4d30F68F...A70F. Nothing to do.
ServiceRegistryTokenUtility 0x3d77596b...1e44 already has drainer() == 0x4d30F68F...A70F. Nothing to do.
```

That run also surfaced a defect in the script itself: `cast wallet address` fails into an **empty string** when the ledger is disconnected, and the run continued with an empty signer (`Casting from: ` blank), only failing later at the owner comparison with a misleading "Signer  is not the current owner". It was harmless only because both contracts short-circuited. Both scripts now validate the derived signer and abort with a clear message - re-tested with the ledger still disconnected, exits 1 immediately.

## 3. `deploy_06a` / `06b` no longer build `$execCmd` as a string

@dagacha's accepted follow-up from #325. Safe today because their arguments - two addresses and a `bytes32` - cannot contain spaces. Worth closing anyway: `deploy_01` shipped a mis-deployment from exactly this pattern, and forge accepts a surplus argument list **silently** whenever the consumed prefix type-checks, so the failure mode is not bounded by argument count.

## Still open, deliberately

- **`deploy_16` has no version pin** - a fleet-wide decision, not a Robinhood one.
- **`ownable_owners.csv`** is regenerated with the handover, not here.
- The handover itself: six `changeOwner` calls, two of which now have `script_l2_10`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/configuration.json
+++ b/docs/configuration.json
@@ -759,7 +759,7 @@
       {
         "name": "IdentityRegistryBridger",
         "artifact": "abis/0.8.30/IdentityRegistryBridger.json",
-        "address": "0x397125902ED2cA2d42104F621f448A2cE1bC8Fb7"
+        "address": "0xa4799B083E0068732456EF45ff9fe5c683658327"
       },
       {
         "name": "IdentityRegistryBridgerProxy",
--- a/scripts/deployment/l2/script_l2_09_registries_change_drainer.sh
+++ b/scripts/deployment/l2/script_l2_09_registries_change_drainer.sh
@@ -59,11 +59,21 @@ if [ "$targetAddress" == "0x0000000000000000000000000000000000000000" ]; then
   exit 1
 fi
 
-for pair in "ServiceRegistryL2:$serviceRegistryAddress" \
-            "ServiceRegistryTokenUtility:$serviceRegistryTokenUtilityAddress"; do
+# Shape-check every address taken from globals, not just null/empty. jq -r preserves a stray
+# trailing space, which passes a null/empty/zero test and then silently defeats every string
+# comparison below: the idempotency branch can never match, while castCmd re-splits on expansion
+# so the transaction itself is correct. A completed handover would then be reported as a hard
+# failure telling the operator to find a derivation path for the bridge mediator.
+for pair in "bridgeMediatorAddress:$targetAddress" \
+            "serviceRegistryAddress:$serviceRegistryAddress" \
+            "serviceRegistryTokenUtilityAddress:$serviceRegistryTokenUtilityAddress"; do
   key="${pair%%:*}"; val="${pair#*:}"
   if [ "$val" == "null" ] || [ -z "$val" ]; then
-    echo "${red}!!! address for $key is not set in $globals${reset}"
+    echo "${red}!!! $key is not set in $globals${reset}"
+    exit 1
+  fi
+  if ! [[ "$val" =~ ^0x[0-9a-fA-F]{40}$ ]]; then
+    echo "${red}!!! $key in $globals is not a well-formed address: '$val'${reset}"
     exit 1
   fi
 done
--- a/scripts/deployment/l2/script_l2_09_registries_change_drainer.sh
+++ b/scripts/deployment/l2/script_l2_09_registries_change_drainer.sh
@@ -1,0 +1,150 @@
+#!/bin/bash
+
+# Sets the drainer of both service registries to the chain's governance control point, by calling changeDrainer(bridgeMediator). The drainer is the destination for slashed operator bonds - ServiceRegistryL2.slashedFunds (native) and ServiceRegistryTokenUtility.mapSlashedFunds[token] (ERC-20). While it is zero both sweep paths are closed: ServiceRegistryL2.drain() requires msg.sender == drainer, and ServiceRegistryTokenUtility.drain(token) reverts ZeroAddress().
+#
+# Covers ServiceRegistryL2 and ServiceRegistryTokenUtility, the two contracts the
+# shell deployment route never had a script for: they are reachable only through
+# the hardhat deploy_11_12_change_drainers.js, so on a shell-route chain the call had to be
+# made by hand with cast send. That is how it was done on chain 4663.
+#
+# The script is idempotent per contract: it skips one already set to the target,
+# and otherwise pre-checks that the signer derived from $derivationPath is the
+# current owner (both setters are owner-gated and would revert with OwnerOnly).
+#
+# Run this BEFORE the ownership handover: changeDrainer is owner-gated, so once
+# owner() is the bridge mediator this becomes a governance proposal crossing the
+# bridge rather than one transaction. Steps 11-12 of docs/deploymentL2.md precede
+# steps 13-15 for exactly this reason.
+
+# Check if $1 is provided
+if [ -z "$1" ]; then
+  echo "Usage: $0 <network>"
+  echo "Example: $0 gnosis_mainnet"
+  exit 1
+fi
+
+red=$(tput setaf 1)
+green=$(tput setaf 2)
+reset=$(tput sgr0)
+
+# Get globals file
+globals="$(dirname "$0")/globals_$1.json"
+if [ ! -f $globals ]; then
+  echo "${red}!!! $globals is not found${reset}"
+  exit 1
+fi
+
+# Read variables using jq
+useLedger=$(jq -r '.useLedger' $globals)
+derivationPath=$(jq -r '.derivationPath' $globals)
+chainId=$(jq -r '.chainId' $globals)
+networkURL=$(jq -r '.networkURL' $globals)
+
+serviceRegistryAddress=$(jq -r '.serviceRegistryAddress' $globals)
+serviceRegistryTokenUtilityAddress=$(jq -r '.serviceRegistryTokenUtilityAddress' $globals)
+bridgeMediatorAddress=$(jq -r '.bridgeMediatorAddress' $globals)
+
+# Target: the chain's governance control point — the bridge mediator on L2, or the
+# L1-aliased Timelock on Arbitrum — stored as bridgeMediatorAddress. Required, with
+# no fallback: timelockAddress is the non-aliased L1 Timelock, an address nobody
+# controls on an L2.
+if [ "$bridgeMediatorAddress" == "null" ] || [ -z "$bridgeMediatorAddress" ]; then
+  echo "${red}!!! bridgeMediatorAddress is not set in $globals${reset}"
+  exit 1
+fi
+targetAddress="$bridgeMediatorAddress"
+
+if [ "$targetAddress" == "0x0000000000000000000000000000000000000000" ]; then
+  echo "${red}!!! targetAddress is the zero address — check $globals${reset}"
+  exit 1
+fi
+
+for pair in "ServiceRegistryL2:$serviceRegistryAddress" \
+            "ServiceRegistryTokenUtility:$serviceRegistryTokenUtilityAddress"; do
+  key="${pair%%:*}"; val="${pair#*:}"
+  if [ "$val" == "null" ] || [ -z "$val" ]; then
+    echo "${red}!!! address for $key is not set in $globals${reset}"
+    exit 1
+  fi
+done
+
+# Check for Alchemy keys
+if [[ "$networkURL" == *"alchemy.com"* ]]; then
+  case $chainId in
+    1)        API_KEY=$ALCHEMY_API_KEY_MAINNET; keyName="ALCHEMY_API_KEY_MAINNET" ;;
+    11155111) API_KEY=$ALCHEMY_API_KEY_SEPOLIA; keyName="ALCHEMY_API_KEY_SEPOLIA" ;;
+    137)      API_KEY=$ALCHEMY_API_KEY_MATIC;   keyName="ALCHEMY_API_KEY_MATIC" ;;
+    80002)    API_KEY=$ALCHEMY_API_KEY_AMOY;    keyName="ALCHEMY_API_KEY_AMOY" ;;
+  esac
+  if [ -n "$keyName" ] && [ "$API_KEY" == "" ]; then
+    echo "set $keyName env variable"
+    exit 1
+  fi
+fi
+
+# Get deployer based on the ledger flag
+if [ "$useLedger" == "true" ]; then
+  walletArgs="-l --mnemonic-derivation-path $derivationPath"
+  deployer=$(cast wallet address $walletArgs)
+else
+  echo "Using PRIVATE_KEY: ${PRIVATE_KEY:0:6}..."
+  walletArgs="--private-key $PRIVATE_KEY"
+  deployer=$(cast wallet address $walletArgs)
+fi
+
+# cast wallet address fails silently into an empty string when the ledger is not
+# connected. Without this the run continues with an empty signer and only fails
+# later, at the owner comparison, with a misleading message.
+if ! [[ "$deployer" =~ ^0x[0-9a-fA-F]{40}$ ]]; then
+  echo "${red}!!! Failed to derive the signer address (is the ledger connected and unlocked?)${reset}"
+  exit 1
+fi
+deployerLc=$(echo "$deployer" | tr '[:upper:]' '[:lower:]')
+targetLc=$(echo "$targetAddress" | tr '[:upper:]' '[:lower:]')
+
+echo "${green}Casting from: $deployer${reset}"
+echo "RPC: $networkURL"
+
+castSendHeader="cast send --rpc-url $networkURL$API_KEY $walletArgs"
+
+for pair in "ServiceRegistryL2:$serviceRegistryAddress" \
+            "ServiceRegistryTokenUtility:$serviceRegistryTokenUtilityAddress"; do
+  name="${pair%%:*}"; addr="${pair#*:}"
+
+  # Addresses are lowercased before comparison because the globals JSON may not be
+  # EIP-55 checksummed; tr is used instead of the ${var,,} expansion so the script
+  # also runs on bash 3.2 (macOS default).
+  current=$(cast call --rpc-url $networkURL$API_KEY $addr "drainer()(address)")
+  if ! [[ "$current" =~ ^0x[0-9a-fA-F]{40}$ ]]; then
+    echo "${red}!!! Failed to read $name drainer() (got: $current)${reset}"
+    exit 1
+  fi
+  currentLc=$(echo "$current" | tr '[:upper:]' '[:lower:]')
+
+  if [ "$currentLc" == "$targetLc" ]; then
+    echo "${green}$name $addr already has drainer() == $targetAddress. Nothing to do.${reset}"
+    continue
+  fi
+
+  # The signer must be the current owner; otherwise changeDrainer reverts with OwnerOnly.
+  currentOwner=$(cast call --rpc-url $networkURL$API_KEY $addr "owner()(address)")
+  currentOwnerLc=$(echo "$currentOwner" | tr '[:upper:]' '[:lower:]')
+  if [ "$currentOwnerLc" != "$deployerLc" ]; then
+    echo "${red}!!! Signer $deployer is not the current $name owner ($currentOwner).${reset}"
+    echo "${red}    Set derivationPath in $globals to the path that controls $currentOwner, then re-run.${reset}"
+    exit 1
+  fi
+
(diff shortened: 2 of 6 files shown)


docs/configuration.json
```diff
@@ -759,7 +759,7 @@
       {
         "name": "IdentityRegistryBridger",
         "artifact": "abis/0.8.30/IdentityRegistryBridger.json",
-        "address": "0x397125902ED2cA2d42104F621f448A2cE1bC8Fb7"
+        "address": "0xa4799B083E0068732456EF45ff9fe5c683658327"
```
dagacha, line 762: Verified against the live Robinhood proxy: `getImplementation()` on `0xE49CB081e8d96920C38aA7AB90cb0294ab4Bc8EA` returns this address. Its `serviceManager()` returns `0x63e66d7ad413C01A7b49C7FF4e3Bb765C4E4bd1b`, matching the manager proxy in the globals. This correction agrees with both the configuration and live deployment.

scripts/deployment/l2/script_l2_09_registries_change_drainer.sh
```diff
@@ -0,0 +1,150 @@
...
+  currentOwner=$(cast call --rpc-url $networkURL$API_KEY $addr "owner()(address)")
+  currentOwnerLc=$(echo "$currentOwner" | tr '[:upper:]' '[:lower:]')
+  if [ "$currentOwnerLc" != "$deployerLc" ]; then
+    echo "${red}!!! Signer $deployer is not the current $name owner ($currentOwner).${reset}"
+    echo "${red}    Set derivationPath in $globals to the path that controls $currentOwner, then re-run.${reset}"
+    exit 1
+  fi
+
+  echo "${green}$name changeDrainer: $current -> $targetAddress${reset}"
+  castArgs="$addr changeDrainer(address) $targetAddress"
```
dagacha, line 149: Validated the transaction path with mocked tools: both registries receive `changeDrainer(bridgeMediatorAddress)` on success, an already-configured registry is skipped independently, and a failed send or status-0 receipt stops the script with exit 1 before proceeding to the next registry. The signer, read, zero-target, missing-address, and wrong-owner failure cases also behaved as expected. No issue found in these checks.

scripts/deployment/l2/script_l2_10_registries_change_owner.sh
```diff
@@ -0,0 +1,149 @@
...
+  currentOwner=$(cast call --rpc-url $networkURL$API_KEY $addr "owner()(address)")
+  currentOwnerLc=$(echo "$currentOwner" | tr '[:upper:]' '[:lower:]')
+  if [ "$currentOwnerLc" != "$deployerLc" ]; then
+    echo "${red}!!! Signer $deployer is not the current $name owner ($currentOwner).${reset}"
+    echo "${red}    Set derivationPath in $globals to the path that controls $currentOwner, then re-run.${reset}"
+    exit 1
+  fi
+
+  echo "${green}$name changeOwner: $current -> $targetAddress${reset}"
+  castArgs="$addr changeOwner(address) $targetAddress"
```
dagacha, line 165: Validated the ownership path with mocked tools: both registries receive `changeOwner(bridgeMediatorAddress)` on success, already-transferred ownership is skipped case-insensitively, and partial completion correctly sends only the remaining transaction. Wrong-owner and transaction-failure cases abort with exit 1. The documented requirement to complete owner-gated setup before this script matches the contracts’ access controls.

dagacha (commented):
No actionable findings at `ec0f176bca807fbc752cb229df003defafefd3f9`.

Review performed with **Codex**, using **gpt-6-astra** at **medium** reasoning effort, confirmed from this review session’s metadata.

Reviewed all five changed files against the surrounding scripts and contract access controls. Independently confirmed the corrected bridger implementation through read-only Robinhood RPC calls. The new scripts call the correct owner-gated setters and document the required drainer-before-ownership order.

Validation: Bash syntax checks passed for all four scripts. An isolated mock harness passed 24 cases: ten scenarios per handover script (success, already configured, partially completed, wrong owner, invalid signer, failed read, send failure, reverted receipt, zero target, and missing registry address), plus both wallet modes for each deployment script. The deployment checks verified exact constructor argument boundaries and the resulting globals update.

Limitations: transaction paths used mocked `cast`/`forge`; no transactions were broadcast and no hardware-wallet signing or end-to-end deployment was performed. The Solidity test suite was not run because this PR changes scripts and configuration only.

dagacha (approved):

scripts/deployment/l2/script_l2_10_registries_change_owner.sh
```diff
@@ -0,0 +1,149 @@
...
+    exit 1
+  fi
+  currentLc=$(echo "$current" | tr '[:upper:]' '[:lower:]')
+
+  if [ "$currentLc" == "$targetLc" ]; then
+    echo "${green}$name $addr already has owner() == $targetAddress. Nothing to do.${reset}"
+    continue
+  fi
+
+  # The signer must be the current owner; otherwise changeOwner reverts with OwnerOnly.
```
OjusWiZard: The ordering constraint the header states at `:14-16` is not enforced anywhere in the loop. This block pre-checks the signer; nothing pre-checks that the drainer is set.

Run `_10` before `_09` on a fresh chain and it exits 0, green, having permanently closed both sweep paths:

```solidity
// ServiceRegistryL2.drain()
if (msg.sender != drainer) revert ManagerOnly(msg.sender, drainer);   // drainer == address(0)

// ServiceRegistryTokenUtility.drain(token)
if (drainer == address(0)) revert ZeroAddress();
```

Both registries are now owned by the bridge mediator with `drainer` unset, so restoring the sweep path needs a governance proposal crossing the bridge. `_09` run afterwards correctly exits 1, but the EOA route is already gone.

The numbering makes this an easy mistake - `scripts/deployment/l2/README.md` still points operators at `deploy_13_15_change_ownerships.js`, so the shell pair is discoverable mainly by reading the directory.

The script already has the primitive one line away:

```sh
currentDrainer=$(cast call --rpc-url $networkURL$API_KEY $addr "drainer()(address)")
if [ "$(echo "$currentDrainer" | tr '[:upper:]' '[:lower:]')" == "0x0000000000000000000000000000000000000000" ]; then
  echo "${red}!!! $name drainer() is still unset - run script_l2_09 first; after this handover it needs a governance proposal.${reset}"
  exit 1
fi
```

scripts/deployment/l2/script_l2_10_registries_change_owner.sh
```diff
@@ -0,0 +1,149 @@
...
+
+# Target: the chain's governance control point — the bridge mediator on L2, or the
+# L1-aliased Timelock on Arbitrum — stored as bridgeMediatorAddress. Required, with
+# no fallback: timelockAddress is the non-aliased L1 Timelock, an address nobody
+# controls on an L2.
+if [ "$bridgeMediatorAddress" == "null" ] || [ -z "$bridgeMediatorAddress" ]; then
+  echo "${red}!!! bridgeMediatorAddress is not set in $globals${reset}"
+  exit 1
+fi
+targetAddress="$bridgeMediatorAddress"
```
OjusWiZard, line 55: `targetAddress` gets `null`, empty and zero-address checks but no shape check - unlike `deployer` at `:97` and `current` at `:117`, which both go through `^0x[0-9a-fA-F]{40}$`. Same for `serviceRegistryAddress` and `serviceRegistryTokenUtilityAddress` at `:61-68`.

The consequence is not a malformed send - it is that the guards and the transaction end up disagreeing. With a trailing space in the globals value, which `jq -r` preserves, I ran this repo's own guard sequence:

```
target="0x4d30F68F...A70F "      # one trailing space
  null-guard:  pass
  empty-guard: pass
  zero-guard:  pass
  idempotency: NO MATCH   <- skip branch is unreachable
  argv cast sees: [0xE3607b00...4b50] [changeOwner(address)] [0x4d30F68F...A70F]
```

`castCmd` re-splits on expansion, so `cast` receives the clean address and the transaction is correct - while every comparison in the script saw the padded string. The idempotency branch at `:123` can then never fire.

That matters most on the second run, which the "idempotent per contract" claim invites: `:123` does not match, control falls through to the owner pre-check, `currentOwner` is now the mediator, and the script exits 1 with

> `!!! Signer 0x... is not the current ServiceRegistryL2 owner (0x4d30F68F...A70F).`
> `    Set derivationPath in globals_... to the path that controls 0x4d30F68F...A70F, then re-run.`

A completed, correct handover reported as a hard failure, with an instruction to find a derivation path for the bridge mediator. Applying the same regex you already use for `deployer` and `current` to all three globals addresses at read time closes both this and the malformed-value case. `script_l2_09` has the same gap, where it costs a redundant `changeDrainer` re-send rather than a false alarm.

OjusWiZard (dismissed):
Approving. The `configuration.json` fix is right - it now agrees with the proxy's implementation slot and the globals, and 4663 joins the six chains that were already consistent.

I checked the ordering claim against the contracts rather than the prose, and it holds: `changeDrainer` and `changeOwner` are both owner-gated on `ServiceRegistryL2` and `ServiceRegistryTokenUtility`, so `_09` before `_10` is genuinely required. Both also `revert ZeroAddress()`, so even a bypassed shell guard cannot burn either registry to zero. Live 4663 state matches your account exactly - drainers already at the mediator, owners still the deployer EOA.

Worth crediting explicitly: both scripts `exit 1` where `script_l2_07` `exit 0`s on a missing globals file and on a missing Alchemy key. In `_07` a typo'd network name is a silent no-op that a runbook reads as done; these two fix that.

Two inline comments, both on `script_l2_10`. Neither blocks - but both failure modes are irreversible from the EOA, so they are worth closing before the next mainnet handover.

---

Three smaller things, not worth their own threads:

- `script_l2_09:130` / `script_l2_10:129` - `currentOwner` is the one derived value in either script that is not regex-checked. A transient RPC error there yields `Signer 0x... is not the current ServiceRegistryL2 owner ().` plus an instruction to change `derivationPath` - verbatim the misleading-message failure the header at `_09:95-97` says it eliminated, reached by a different route. In `_10` the call is also redundant: `:116` already read and validated `owner()` into `current`.
- `script_l2_10:16` says "steps 13-15 of docs/deploymentL2.md". Step 15 is the `ServiceManagerToken` transfer, which is `script_l2_07`; `_10` is steps 13-14. `_09`'s reference to 11-12 is accurate.
- `deploy_06a:36` / `deploy_06b:35` - `constructorArgs` no longer feeds the deploy, but still feeds the `Deployment of:` log at `:51`/`:50` and the `cast abi-encode` at `:76`/`:75`. If the constructor gains or reorders an argument, the natural edit is that variable: the contract would deploy with one set, the log announce another, and `forge verify-contract` submit encoded args that do not match the bytecode. Keeping one array and expanding it `"${constructorArgs[@]}"` in all three places would close it.

One separate note, since this PR cites those lines as its authority for the ordering: `docs/deploymentL2.md:13-14` describes steps 11-12 as changing the drainer "calling `changeManager(BridgeMediator)`". Those are different functions - `changeManager` sets `manager` (`GenericRegistry.sol:54`), `changeDrainer` sets `drainer` (`ServiceRegistryL2.sol:109`). `script_l2_09` implements the step correctly; the doc line would point `manager` at the bridge mediator if followed literally. One-line fix, and this is the natural PR for it.

For the record on something I raised on #325 and want to correct forward: `operatorWhitelistAddress` equalling `timelockAddress` in `globals_robinhood_mainnet.json` is not a defect. `0x3C1fF68f...` carries 2279 B of code on 4663 as a real `OperatorWhitelist`, and coincides with the L1 Timelock only through deployer-nonce alignment. `configuration.json` agrees. Worth knowing before someone "fixes" it.

dagacha (commented):
Reviewed with **OpenAI Codex (GPT-6)**. The exact model variant and reasoning-effort setting are not exposed in this session.

Reviewed commit `ec0f176bca807fbc752cb229df003defafefd3f9`. **No actionable findings.**

The new registry calls match the contracts’ access controls and the existing Hardhat deployment flow. All four changed shell scripts passed `bash -n`, and both new scripts passed mocked success, no-op, signer failure, read failure, wrong-owner, send failure, and reverted-receipt scenarios (14 scenarios total). `git diff --check` also passed.

Read-only RPC verification against Robinhood chain **4663** confirmed:
- `IdentityRegistryBridgerProxy.getImplementation()` returns `0xa4799B083E0068732456EF45ff9fe5c683658327`, matching the corrected configuration entry.
- The live implementation’s `serviceManager()` returns `0x63e66d7ad413C01A7b49C7FF4e3Bb765C4E4bd1b`, matching the configured ServiceManager proxy.
- The replaced implementation (`0x397125902ED2cA2d42104F621f448A2cE1bC8Fb7`) returns the zero address for `serviceManager()`.
- Both registries’ `drainer()` values are `0x4d30F68F5AA342d296d4deE4bB1Cacca912dA70F`, matching the configured governance mediator.
- Both registries’ `owner()` values remain `0xEB2A22b27C7Ad5eeE424Fd90b376c745E60f914E`, consistent with the pending ownership handover.

No transactions were sent. Ledger signing and live transaction execution were not tested. There are no line-specific findings requiring inline comments.

scripts/deployment/l2/script_l2_10_registries_change_owner.sh
```diff
@@ -0,0 +1,149 @@
...
+  fi
+  currentLc=$(echo "$current" | tr '[:upper:]' '[:lower:]')
+
+  if [ "$currentLc" == "$targetLc" ]; then
+    echo "${green}$name $addr already has owner() == $targetAddress. Nothing to do.${reset}"
+    continue
+  fi
+
+  # The signer must be the current owner; otherwise changeOwner reverts with OwnerOnly.
+  currentOwner=$(cast call --rpc-url $networkURL$API_KEY $addr "owner()(address)")
```
dagacha: The ordering constraint the header states is not enforced anywhere in this loop. This block pre-checks the signer; nothing pre-checks that `drainer()` is set.

On a fresh chain, `_10` then `_09`: `_10` exits 0 having moved both `owner()`s to the alias; `_09` then exits 1 (`OwnerOnly`) and `drainer()` stays `0`. `ServiceRegistryL2.drain()` requires `msg.sender == drainer`; `ServiceRegistryTokenUtility.drain(token)` reverts `ZeroAddress()` while `drainer == 0`. Restoring either sweep path needs a bridged `changeDrainer`.

4663 is already past this (drainers set). The next shell-route chain is not — README still points operators at `deploy_13_15_change_ownerships.js`.

Abort here if `drainer()` is unset (or not the target), before `changeOwner`:

```sh
currentDrainer=$(cast call --rpc-url $networkURL$API_KEY $addr "drainer()(address)")
if [ "$(echo "$currentDrainer" | tr '[:upper:]' '[:lower:]')" == "0x0000000000000000000000000000000000000000" ]; then
  echo "${red}!!! $name drainer() is still unset - run script_l2_09 first; after this handover it needs a governance proposal.${reset}"
  exit 1
fi
```

dagacha (changes requested):
Config fix checks out on 4663: proxy `getImplementation()` is `0xa4799B08…`, matching globals; `0x39712590…` is the abandoned impl (`serviceManager() == 0`).

**Blocker:** `script_l2_10` does not enforce `script_l2_09`. Header says run drainers first; the loop never reads `drainer()`. On a fresh chain, `_10` exits 0 having moved both `owner()`s to the alias; `_09` then exits 1 (`OwnerOnly`) and `drainer()` stays `0`. `ServiceRegistryL2.drain` / `TokenUtility.drain` are then closed until a bridged `changeDrainer`. 4663 is already past this; the next shell-route chain is not.

Abort `_10` if either `drainer()` is unset, before the first `changeOwner`.

Review performed with **omp**, using **grok-4.6** at **medium** reasoning effort.

dagacha (commented):
Reviewed the landed merge (`ad202b7`, including `0408c15`) against RPC chain 4663.

No actionable findings. The bridger entry matches the live proxy implementation (`getImplementation()` = `0xa4799B08…`, `serviceManager()` = ServiceManagerProxy). `script_l2_10` now aborts when `drainer()` is unset before `changeOwner`. `deploy_06a` / `06b` no longer build `$execCmd` as a string.

Reviewed with omp. Main LLM: grok-4.6. Effort: medium.

scripts/deployment/l2/script_l2_10_registries_change_owner.sh
```diff
@@ -0,0 +1,176 @@
...
+  # while drainer() is unset permanently closes both sweep paths from the EOA:
+  # ServiceRegistryL2.drain() requires msg.sender == drainer, and
+  # ServiceRegistryTokenUtility.drain(token) reverts ZeroAddress(). Running _09 afterwards correctly
+  # fails OwnerOnly, by which point restoring the sweep needs a governance proposal over the bridge.
+  currentDrainer=$(cast call --rpc-url $networkURL$API_KEY $addr "drainer()(address)")
+  if ! [[ "$currentDrainer" =~ ^0x[0-9a-fA-F]{40}$ ]]; then
+    echo "${red}!!! Failed to read $name drainer() (got: $currentDrainer)${reset}"
+    exit 1
+  fi
+  if [ "$(echo "$currentDrainer" | tr '[:upper:]' '[:lower:]')" == "0x0000000000000000000000000000000000000000" ]; then
```
dagacha, line 149: [P2] Require the intended drainer before transferring ownership

This guard only rejects the zero address. If either registry has an old or mistakenly configured nonzero drainer, the script still transfers ownership to the bridge mediator, leaving native draining authority (ServiceRegistryL2) or the destination for ERC-20 proceeds (ServiceRegistryTokenUtility) assigned to that other address. Correcting it afterwards requires governance because changeDrainer is owner-gated. In a mocked cast run with drainer() set to the deployer instead of bridgeMediatorAddress, the script sent both ownership transactions and exited 0. Compare the normalized currentDrainer against $targetLc and abort on any mismatch, directing the operator to run script_l2_09 first.

dagacha (commented):
Review performed with Codex. LLM: GPT-6. Effort: Medium.

Found one actionable issue, detailed inline. Validated with mocked cast executions and bash syntax checks; no live transactions were sent.

PR #327, opened 2026-09-10
fix: require the intended drainer before handing over ownership

DavidMinarsch:
The drainer guard added in #326 only rejected the **zero** address. A stale or mistyped non-zero drainer survived the handover, leaving native draining authority (`ServiceRegistryL2`) or the ERC-20 proceeds destination (`ServiceRegistryTokenUtility`) assigned to that address — correctable only by governance, since `changeDrainer` is owner-gated.

@dagacha reproduced it: with `drainer()` set to the deployer, the script sent both ownership transactions and exited 0.

It now compares the normalised drainer against the intended target and aborts on any mismatch, distinguishing "still unset" from "set to something else" so the message tells the operator which case they're in.

Fixtures:

```
intended mediator                exit=0
unset                            exit=1  drainer() is still unset.
deployer (the reported case)     exit=1  drainer() is 0xEB2A22b2..., not the intended 0x4d30F68F...
unreadable                       exit=1  Failed to read drainer()
```

4663 is already past this point — both drainers are the mediator — so this protects the next shell-route chain.

### DIFF
--- a/scripts/deployment/l2/script_l2_10_registries_change_owner.sh
+++ b/scripts/deployment/l2/script_l2_10_registries_change_owner.sh
@@ -146,9 +146,18 @@ for pair in "ServiceRegistryL2:$serviceRegistryAddress" \
     echo "${red}!!! Failed to read $name drainer() (got: $currentDrainer)${reset}"
     exit 1
   fi
-  if [ "$(echo "$currentDrainer" | tr '[:upper:]' '[:lower:]')" == "0x0000000000000000000000000000000000000000" ]; then
-    echo "${red}!!! $name drainer() is still unset. Run script_l2_09_registries_change_drainer.sh first${reset}"
-    echo "${red}    — after this handover, setting it needs a governance proposal over the bridge.${reset}"
+  # Require the intended drainer, not merely a non-zero one. A stale or mistyped value would otherwise
+  # survive the handover, leaving native draining authority (ServiceRegistryL2) or the ERC-20 proceeds
+  # destination (ServiceRegistryTokenUtility) with that address, correctable only by governance.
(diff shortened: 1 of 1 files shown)


dagacha (commented):
No actionable findings.

`script_l2_10` now aborts when `drainer()` is not the intended target, not merely when it is zero. Distinct messages for unset vs stale/mistyped. That is the send path that previously exited 0 with `drainer() == deployer`.

Reviewed with omp. Main LLM: grok-4.6. Effort: medium.

dagacha (commented):
Review performed with Codex (LLM: GPT-6; effort: Medium).

No actionable findings in the reviewed changes. The guard requires the intended drainer before transferring each registry's ownership, with case-insensitive address comparison consistent with the drainer setup script.

Validation: Bash syntax check and nine stubbed cast scenarios passed, covering intended and differently cased addresses, unset/incorrect/unreadable drainers for each registry, and already-completed ownership handovers. Checks are per contract: a failure on the second registry can occur after the first registry's ownership has transferred. No live transactions were sent.

No inline comments were added because no actionable issues were identified.

OjusWiZard (approved):
Approving. I ran this revision of `script_l2_10` against a stubbed `cast`:

```
drainer = intended mediator            exit=0  sends=2
drainer unset                          exit=1  sends=0
drainer = deployer (the reported case) exit=1  sends=0
drainer unreadable                     exit=1  sends=0
drainer = mediator, lowercase          exit=0  sends=2
target with trailing space / malformed exit=1  sends=0
```

The rule it now enforces - drainer equals the bridge mediator - matches live state on every production L2: Arbitrum, Base, Celo, Gnosis, Mode, Optimism, Polygon and Robinhood.

PR #329, opened 2026-09-10
fix(audit): 4663 entries named artifacts that were never built for it

DavidMinarsch:
Eleven of the fifteen registries entries for chain 4663 in `docs/configuration.json` named artifacts that do not match the deployed bytecode. The audit's Tier 1 length check passed on every one — the lengths coincide — so only the trailer check saw it.

## Why they don't match

Nothing in the deployment produces these files. `deploy_*.sh` run `forge create --broadcast`, which compiles from source and broadcasts without writing an artifact back into the repo. So an entry matches only if someone has committed a build made the same way the contract was deployed. The 4663 entries were picked by contract name, and that went wrong in two independent ways:

- **Seven** pointed into `abis/0.8.28/`, which holds **hardhat** artifacts from 2024-10-14, for contracts deployed with **solc 0.8.30** through forge. Wrong compiler and wrong toolchain — and the toolchain matters on its own, because the metadata IPFS hash is taken over the metadata JSON, which records source paths and remappings.
- **Four more** pointed at `abis/0.8.30/` files that are forge builds but predate the source they were matched against; `RecoveryModule.json` dates from 2025-06-25.

The four entries that *did* match are exactly the ones whose `abis/0.8.30/` artifact is a current forge build of the same source, which is the pattern that has to hold generally.

## The fix

Ten entries now point at `abis/deployed/Robinhood*.json` — forge builds that reproduce the deployed bytecode, trailer included, at the repo's own settings (solc 0.8.30, `optimizer_runs = 750`, `evm_version = prague`). Four entries are unchanged because they already matched. Nothing under `abis/<solc>/` is modified, and no other chain is touched.

`StakingToken` keeps `abis/0.8.28/StakingToken.json` and stays the one warn on this chain. Its deployed body is byte-identical to that artifact — zero differing bytes — but the artifact came from the repo's hardhat era, so no forge build reproduces its metadata hash. `abis/README.md` now records the settings that do reproduce the body (solc 0.8.28, optimizer 750, `evm_version = cancun`, source at `5efeb59`) so it doesn't have to be worked out again. Sourcify records the same contract as `match` rather than `exact_match`, for the same reason.

## Verification

Checked per entry against chain 4663 — every configured artifact resolved and compared to `eth_getCode`:

| | before | after |
|---|---|---|
| artifact agrees on length **and** trailer | 4 / 15 | **14 / 15** |

The remaining one is `StakingToken`, explained above.

Found while auditing the §3.7 ownership handover. My defect, from the 4663 entry added in #325.


dagacha (commented):
No blocking issues.

Live 4663: all 15 `configuration.json` artifacts match `eth_getCode` on length; 14/15 match the 43-byte trailer. The ten `abis/deployed/Robinhood*.json` files equal on-chain code once `immutableReferences` ranges are masked. Both proxy entries still take ABI from `configContracts[i-1]` (`ServiceManager`, `IdentityRegistryBridger`).

Review performed with **omp**, using **grok-4.6** at **medium** reasoning effort.

OjusWiZard (approved):
Approving. I compared every 4663 entry at this head with the deployed code:

- **14 of 15 agree on length and metadata trailer**, with no differing bytes outside immutable slots. `StakingToken` is the one exception: its body is byte-identical and only the trailer differs.
- **The ten new `abis/deployed/Robinhood*.json`** record solc `0.8.30+commit.73712a01`, 750 runs and `prague`, in the `deployedBytecode.object` shape the audit reads.
- **The four entries left on `abis/0.8.30/`** already matched, as stated.

